PR #1, opened 2019-09-24
Make sure valid URL is constructed

britter:
Several customers ran into the problem that the script created an
invalid query URL because they configured the Gradle Enterprise server
URL in gradle-enterprise.xml with a trailing slash. This cuased the
query URL to contain an additional slash after the server base URL which
does not work.
This fix simply makes sure that a slash is only added if it's not
already there.


britter:
@facewindu done

britter:
@facewindu implemented your suggestions.

PR #2, opened 2019-09-24
Don't add search links unless server is configured

marcphilipp:
Some customers don't have the server configured by default which lead
to exceptions because of invalid URLs.


autonomousapps:
FWIW, I assumed (wrongly, obviously) that people would actually look at the script before applying it, and then test it, and then modify it as they saw fit.

autonomousapps:
I agree with @facewindu

marcphilipp:
Done.

@facewindu PTAL!

marcphilipp:
@facewindu Can you please merge this? I don't have permission.

marcphilipp:
@facewindu Thanks! 👍

PR #4, opened 2019-11-06
Add script that adds a custom value for each Maven reactor project

marcphilipp:
This allows for filtering by selected project in the build scans list.

### DIFF
--- a/maven-reactor-projects/maven-reactor-projects.groovy
+++ b/maven-reactor-projects/maven-reactor-projects.groovy
@@ -1,0 +1,11 @@
+/**
+ * This Groovy script adds a custom value for each Maven project in the reactor.
+ */
+
+def buildScan = session.lookup('com.gradle.maven.extension.api.scan.BuildScanApi')
+
+buildScan.executeOnce('maven-reactor-projects') { api ->
+    session.projects.each {
+        api.value('project', it.name)
+    }
+}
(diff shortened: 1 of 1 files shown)


maven-reactor-projects/maven-reactor-projects.groovy
```diff
@@ -0,0 +1,11 @@
+/**
+ * This Groovy script adds a custom value for each Maven project in the reactor.
+ */
+
+def buildScan = session.lookup('com.gradle.maven.extension.api.scan.BuildScanApi')
+
+buildScan.executeOnce('maven-reactor-projects') { api ->
```
oehme: ❌ This is only needed for things that are global. If you want something per project, you can just access the project directly, since this script is already executed for each project.

marcphilipp: Isn't `session.projects` global? Would you instead want to add sth. like the following?
```groovy
if (session.projects.contains(project)) {
	api.value('project', project.name)
}
```

marcphilipp: `buildScan.value('project', project.name)` is sufficient. 👍

marcphilipp:
@facewindu Could you please merge this? @etiennestuder has given me his ok to collect such snippets here.

marcphilipp:
Thanks!

maven-reactor-projects/maven-reactor-projects.groovy
```diff
@@ -0,0 +1,7 @@
+/**
+ * This Groovy script adds a custom value for each Maven project in the reactor.
+ */
+
+def buildScan = session.lookup('com.gradle.maven.extension.api.scan.BuildScanApi')
+
+buildScan.value('project', project.name)
```
oehme: ❌ This will override the value for every project. Sorry I missed that, it was also true for the earlier version of the script.

PR #11, opened 2020-05-18
Add workaround for classloading issue

marcphilipp:
Due to the classloading issue described in
https://issues.apache.org/jira/browse/MNG-6906 we now consume
`BuildScanApi` in a separate class to avoid classpath issues. The
lookup and workaround are encapsulated in `BuildScanApiAccessor`.


etiennestuder:
Thanks, @marcphilipp

PR #20, opened 2020-06-15
Use TC temp dir for storage of GE Maven extension

marcphilipp:
To avoid polluting `~/.m2`, the temp dir provided by TeamCity is now used
for temporary storage by the Gradle Enterprise Maven extension used in
the build.

### DIFF
--- a/.teamcity/settings.kts
+++ b/.teamcity/settings.kts
@@ -120,7 +120,7 @@ fun BuildSteps.maven(goal: String, args: String?) {
     this.maven {
         pomLocation = "common-custom-user-data-maven-extension/pom.xml"
         goals = goal
-        runnerArgs = "-Dgradle.enterprise.url=https://e.grdev.net" + (args?.let { " $it" } ?: "")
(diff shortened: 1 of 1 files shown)


.teamcity/settings.kts
```diff
@@ -120,7 +120,7 @@ fun BuildSteps.maven(goal: String, args: String?) {
     this.maven {
         pomLocation = "common-custom-user-data-maven-extension/pom.xml"
         goals = goal
-        runnerArgs = "-Dgradle.enterprise.url=https://e.grdev.net" + (args?.let { " $it" } ?: "")
+        runnerArgs = "-Dgradle.enterprise.url=https://e.grdev.net -Dgradle.enterprise.storage.directory=%teamcity.build.tempDir%/.gradle-enterprise" + (args?.let { " $it" } ?: "")
```
snoopcheri, line 123: Aha! So `gradle.enterprise.storage.directory` is a configuration parameter that our Gradle Maven extension supports. And that directory basically stores the equivalent of what the GE Gradle plugin stores under `~/.gradle/enterprise` by default.

marcphilipp: Yep: https://docs.gradle.com/enterprise/maven-extension/#anatomy_of_the_gradle_enterprise_directory

PR #21, opened 2020-06-15
Optionally evaluate Groovy script if exists

marcphilipp:
In addition to the custom user data added via Java code in
`CustomUserData`, the extension now checks for a Groovy script in
`.mvn/gradle-enterprise-custom-user-data.groovy` and evaluates it if it
exists.

### DIFF
--- a/common-custom-user-data-maven-extension/src/main/java/com/gradle/GroovyScriptUserData.java
+++ b/common-custom-user-data-maven-extension/src/main/java/com/gradle/GroovyScriptUserData.java
@@ -1,0 +1,48 @@
+package com.gradle;
+
+import com.gradle.maven.extension.api.scan.BuildScanApi;
+import groovy.lang.Binding;
+import groovy.lang.GroovyShell;
+import org.apache.maven.MavenExecutionException;
+import org.apache.maven.execution.MavenSession;
+import org.apache.maven.monitor.logging.DefaultLog;
+import org.codehaus.plexus.logging.Logger;
+
+import java.io.File;
+import java.io.IOException;
+
+class GroovyScriptUserData {
+
+    static void addToBuildScan(MavenSession session, BuildScanApi buildScan, Logger logger) throws MavenExecutionException {
+        File scriptFile = getScriptFile(session);
+        if (scriptFile.exists()) {
+            logger.debug("Evaluating custom user data Groovy script: " + scriptFile);
+            evaluateGroovyScript(session, buildScan, logger, scriptFile);
+        } else {
+            logger.debug("Skipping evaluation of custom user data Groovy script because it does not exist: " + scriptFile);
+        }
+    }
+
+    private static File getScriptFile(MavenSession session) {
+        File rootDir = session.getRequest().getMultiModuleProjectDirectory();
+        return new File(rootDir, ".mvn/gradle-enterprise-custom-user-data.groovy");
+    }
+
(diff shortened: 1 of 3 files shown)


common-custom-user-data-maven-extension/src/main/java/com/gradle/GroovyScriptUserData.java
```diff
@@ -0,0 +1,48 @@
...
+            throw new MavenExecutionException("Failed to evaluate custom user data Groovy script: " + scriptFile, e);
+        }
+    }
+
+    private static Binding prepareBinding(MavenSession session, BuildScanApi buildScan, Logger logger) {
+        Binding binding = new Binding();
+        binding.setVariable("project", session.getTopLevelProject());
+        binding.setVariable("session", session);
+        binding.setVariable("buildScan", buildScan);
+        binding.setVariable("log", new DefaultLog(logger));
```
marcphilipp, line 45: This is less than [what the GMavenPlus2 plugin provides](https://github.com/groovy/GMavenPlus/blob/a21bca3b086110d770ee639737edb37b89f34eb6/src/main/java/org/codehaus/gmavenplus/mojo/AbstractToolsMojo.java#L107-L144) but should be sufficient IMHO.

common-custom-user-data-maven-extension/src/main/java/com/gradle/GroovyScriptUserData.java
```diff
@@ -0,0 +1,48 @@
...
+            logger.debug("Evaluating custom user data Groovy script: " + scriptFile);
+            evaluateGroovyScript(session, buildScan, logger, scriptFile);
+        } else {
+            logger.debug("Skipping evaluation of custom user data Groovy script because it does not exist: " + scriptFile);
+        }
+    }
+
+    private static File getScriptFile(MavenSession session) {
+        File rootDir = session.getRequest().getMultiModuleProjectDirectory();
+        return new File(rootDir, ".mvn/gradle-enterprise-custom-user-data.groovy");
```
snoopcheri, line 28: ❓ I guess it's enough when the script filename is fixed. But where/how will we document this to the user?

marcphilipp: We should probably add it [to the docs](https://docs.gradle.com/enterprise/maven-extension/#custom_extension) and the trial navigator.

facewindu: +1 I don't think we need to offer customization of this file

common-custom-user-data-maven-extension/src/main/java/com/gradle/GroovyScriptUserData.java
facewindu: +1 we can always add more later

facewindu: but we should document what is available to the script, of course

marcphilipp:
> 👍 just requested documentation updates to happen

@facewindu You mean what I wrote [here](https://github.com/gradle/gradle-enterprise-build-config-samples/pull/21#discussion_r440652394)?

common-custom-user-data-maven-extension/src/main/java/com/gradle/GroovyScriptUserData.java
facewindu: already covered

PR #22, opened 2020-07-02
Update instructions for custom groovy script

runningcode:


runningcode:
Misunderstanding. This script duplicates the logic which already exists in the extension.

PR #24, opened 2020-07-27
Add script for rolling out maven extension to many repositories

britter:


maven-extension-rollout/rollout.sh
```diff
@@ -0,0 +1,52 @@
+#!/bin/bash
```
marcphilipp, line 1: Should we use `set +e` to make the script stop on the first error?

snoopcheri: Do you mean `set -e`? With `set +e` the script does not automatically stop.

The problem with this is that you cannot do something (like `git checkout .mvn/extensions.xml` below) and check for the exit code, since the script immediately stops.

maven-extension-rollout/rollout.sh
```diff
@@ -0,0 +1,52 @@
...
+
+commit_msg="Apply Gradle Enterprise Maven Extension"
+basedir="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
```
snoopcheri, line 4: I believe `BASH_SOURCE[0]` is not available with bash older than v3. So depending on where the script runs you could replace it with `$0`. OTOH, `BASH_SOURCE[0]` is "better" than `$0`.

maven-extension-rollout/rollout.sh
```diff
@@ -0,0 +1,52 @@
...
+  echo "($i/$num) Updating $line..."
+  process_repository $line
+  ((i++))
+  done < $repositories
+}
+
+function process_repository() {
+  repository_name="${1##*/}"
+  # clone the repository without actually downloading files or history
+  git clone -n $1 $checkout_area/$repository_name --depth 1 >& /dev/null
```
snoopcheri, line 22: I typically put all `$something` into quotes (`"$something"`) because this way it even works when the variable contains spaces.

maven-extension-rollout/rollout.sh
```diff
@@ -0,0 +1,52 @@
...
+  pushd $checkout_area/$repository_name >& /dev/null
+  git reset HEAD . >& /dev/null
+  git checkout .mvn/extensions.xml >& /dev/null
+  if [ $? -eq 0 ]; then
```
snoopcheri, line 26: This will most likely not work with `set -e`.

maven-extension-rollout/rollout.sh
```diff
@@ -0,0 +1,52 @@
+#!/bin/bash
+
+commit_msg="Apply Gradle Enterprise Maven Extension"
+basedir="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
+repositories=$basedir/repositories.txt
+extensions_xml=$basedir/extensions.xml
+checkout_area=`mktemp -d`
+
+function process_repositories() {
+  num=`wc -l $repositories | awk '{ print $1 }'`
```
snoopcheri, line 10: Not sure why `| awk ...` is required. `wc -l` should only print one line anyway?

britter: Here is the output without the `awk`. Not sure why it does it because if seems to be harder to process but that's the way it is 🤷‍♂️ 
```
❯ wc -l repositories.txt
       2 repositories.txt
```

maven-extension-rollout/rollout.sh
britter: Done in d5fdb67

maven-extension-rollout/rollout.sh
snoopcheri: Ah! I usually use `wc -l` with piping stuff into it. Didn't realize that when using it this way it actually writes the filename - which actually makes sense since you could feed it multiple files (eg with`ed -l *.txt)

Issue #25, opened 2020-07-28
Gradle Enterprise Maven Extension rollout script should also append to .gitignore


britter:
When applying the Gradle Enterprise Maven Extension the `.mvn/.gradle-enterprise` should not be committed. The rollout script should therefore append this path to the `.gitignore` file in the repository root.

PR #26, opened 2020-07-28
Enhance rollout.sh

LukasW:
Fixed: https://github.com/gradle/gradle-enterprise-build-config-samples/issues/25
Process maven projects only

### DIFF
--- a/maven-extension-rollout/rollout.sh
+++ b/maven-extension-rollout/rollout.sh
@@ -26,11 +26,19 @@ function process_repository() {
   if [ $? -eq 0 ]; then
     echo ".mvn/extensions.xml already exists in $1, skipping..." >&2
   else
-    mkdir -p .mvn
-    cp "$extensions_xml" .mvn
-    git add .mvn/extensions.xml >& /dev/null
-    git commit -m "$commit_msg" >& /dev/null
-    git push >& /dev/null
+    # Only process maven projects
+    if [ -z "$(git ls-tree -r HEAD --name-only | grep pom.xml)" ]; then  
+      echo "$1 is not a mvn project, skipping..." >&2
+    else 
+      mkdir -p .mvn
+      cp "$extensions_xml" .mvn
+      git add .mvn/extensions.xml >& /dev/null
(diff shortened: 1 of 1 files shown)


maven-extension-rollout/rollout.sh
```diff
@@ -26,11 +26,19 @@ function process_repository() {
...
     echo ".mvn/extensions.xml already exists in $1, skipping..." >&2
   else
-    mkdir -p .mvn
-    cp "$extensions_xml" .mvn
-    git add .mvn/extensions.xml >& /dev/null
-    git commit -m "$commit_msg" >& /dev/null
-    git push >& /dev/null
+    # Only process maven projects
+    if [ -z "$(git ls-tree -r HEAD --name-only | grep pom.xml)" ]; then  
+      echo "$1 is not a mvn project, skipping..." >&2
```
britter, line 31: Nice 👍

maven-extension-rollout/rollout.sh
```diff
@@ -26,11 +26,19 @@ function process_repository() {
...
+    else 
+      mkdir -p .mvn
+      cp "$extensions_xml" .mvn
+      git add .mvn/extensions.xml >& /dev/null
+      echo ".gradle-enterprise/" > .mvn/.gitignore
+      git add .mvn/.gitignore
```
britter, line 37: I think it's more idiomatic to only have a single `.gitignore` in the repository root. How about this?

```suggestion
      git checkout .gitignore >& /dev/null
      echo ".mvn/.gradle-enterprise/" >> .gitignore
      git add .gitignore
```

britter: @LukasW are you okay with this change?

LukasW: That's fine with me. 
Actually, I would prefer the .gitignore inside the .mvn directory, because if you would delete the .mvn directory, the ignoring would be gone as well.

britter: We've discussed this in the team and decided that we're going with the main `.gitignore`. Since the whole project is just a template for people to setup their own automation users are free to customize this to their needs.

PR #27, opened 2020-07-29
Use template directory

LukasW:
**Proposal**

Uses a whole template directory.

Additionally there are two switches:
-p (Push): Only then the stuff is pushed
-f (Force): The .mvn directory will be overwritten


LukasW:
If you like it, I will also update the documentation.

maven-extension-rollout/.mvn/.gitignore
```diff
@@ -0,0 +1 @@
+.gradle-enterprise/
```
britter, line 1: As stated in #26 I think this should do to the main `.gitignore`

maven-extension-rollout/.mvn/extensions.xml
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<extensions>
+    <extension>
+        <groupId>com.gradle</groupId>
+        <artifactId>gradle-enterprise-maven-extension</artifactId>
+        <version>1.5.3</version>
```
britter, line 6: Let's use the latest version.
```suggestion
        <version>1.6</version>
```

maven-extension-rollout/rollout.sh
```diff
@@ -2,9 +2,17 @@
...
-extensions_xml=$basedir/extensions.xml
 checkout_area=`mktemp -d`
+repositories=$basedir/repositories.txt
+
+# process arguments "$1", "$2", ... (i.e. "$@")
+while getopts "pf" opt; do
+    case $opt in
+    p) push=true ;; 
+    f) force=true ;;
+    \?) ;; # Handle error: unknown option or missing required argument.
```
britter, line 13: Would be good to log an unknown option here.

LukasW:
> Nice idea to provide the `force` option. I think we should finish #26 first or does this replace #26?

This request replaces #26

britter:
@LukasW I've merged #26. Can you rebase this on master? I think we need to modify the code that appends to `.gitignore`. If `force=true` we need to check whether `.gitignore` already contains an entry for the gradle-enterprise directory. We only need to add it to the `.gitignore` if not.

PR #30, opened 2020-08-11
Add script to capture execution root name/artifactId

marcphilipp:
This allows to filter the list of build scans by the project Maven was
invoked on, e.g. using `mvn -f`.

### DIFF
--- a/capture-top-level-project/gradle-enterprise-custom-user-data.groovy
+++ b/capture-top-level-project/gradle-enterprise-custom-user-data.groovy
@@ -2,5 +2,5 @@
 // It needs to be placed in .mvn/gradle-enterprise-custom-user-data.groovy in the root project directory.
 // (see https://docs.gradle.com/enterprise/maven-extension/#using_the_common_custom_user_data_maven_extension)
 
-buildScan.value('topLevelProject.name', session.topLevelProject.name)
-buildScan.value('topLevelProject.artifactId', session.topLevelProject.artifactId)
+buildScan.value('executionRoot.name', session.topLevelProject.name)
+buildScan.value('executionRoot.artifactId', session.topLevelProject.artifactId)
--- a/capture-top-level-project/gradle-enterprise-custom-user-data.groovy
+++ b/capture-top-level-project/gradle-enterprise-custom-user-data.groovy
@@ -1,0 +1,6 @@
+// This script is meant to be used with the common-custom-user-data-maven-extension.
+// It needs to be placed in .mvn/gradle-enterprise-custom-user-data.groovy in the root project directory.
+// (see https://docs.gradle.com/enterprise/maven-extension/#using_the_common_custom_user_data_maven_extension)
+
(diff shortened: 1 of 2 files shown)


capture-top-level-project/gradle-enterprise-custom-user-data.groovy
```diff
@@ -0,0 +1,6 @@
+// This script is meant to be used with the common-custom-user-data-maven-extension.
+// It needs to be placed in .mvn/gradle-enterprise-custom-user-data.groovy in the root project directory.
+// (see https://docs.gradle.com/enterprise/maven-extension/#using_the_common_custom_user_data_maven_extension)
+
+buildScan.value('topLevelProject.name', session.topLevelProject.name)
+buildScan.value('topLevelProject.artifactId', session.topLevelProject.artifactId)
```
britter, line 6: ⭕️ Should this also be added to the common custom user data maven extension?

capture-top-level-project/gradle-enterprise-custom-user-data.groovy
```diff
@@ -0,0 +1,6 @@
+// This script is meant to be used with the common-custom-user-data-maven-extension.
+// It needs to be placed in .mvn/gradle-enterprise-custom-user-data.groovy in the root project directory.
+// (see https://docs.gradle.com/enterprise/maven-extension/#using_the_common_custom_user_data_maven_extension)
+
+buildScan.value('topLevelProject.name', session.topLevelProject.name)
```
britter, line 5: ⭕️ Okay, this (maven) is really confusing... to the topLevelProject is the project the build was started with although there might be an enclosing project (the one with the `.mvn` folder). I wonder whether `topLevelProject` is - although technically correct - a good name for this custom value.

marcphilipp: I'm also not sure about the name. I decided to stick with Maven's name for this (which is also inconsistent internally since some places call it "execution root") instead of coming up with a new name.

@etiennestuder WDYT?

capture-top-level-project/gradle-enterprise-custom-user-data.groovy
marcphilipp: I think it should be either here or there. I'm not yet convinced this is useful for all customers so I added it as an "extension".

capture-top-level-project/gradle-enterprise-custom-user-data.groovy
snoopcheri: Wouldn't the "execution root" be the place where you've executed `mvn` - which is not necessarily the project root, if you only build a submodule.

capture-top-level-project/gradle-enterprise-custom-user-data.groovy
marcphilipp, line 6:
```suggestion
buildScan.value('executionRoot.name', session.topLevelProject.name)
buildScan.value('executionRoot.artifactId', session.topLevelProject.artifactId)
```

PR #31, opened 2020-08-12
Remove tag for Gradle and Maven Versions

runningcode:
This is built in to Gradle Enterprise now.


etiennestuder:
@runningcode did you remove it in all places (gradle/maven/extension) ?

runningcode:
@etiennestuder removed them as well. I didn't realize we were also tagging maven versions in this way.

etiennestuder:
LGTM. Thanks @runningcode

PR #33, opened 2020-08-12
Add example tagging of Android Studio Version.

### LINKED ISSUE #7: Tag Android Studio version
Is it possible to get the Android studio version to add it as a tag?
Thanks

runningcode:
Note this will work starting with Android Studio 4.1.
See: https://issuetracker.google.com/issues/152941155

Fixes #7

### DIFF
--- a/common-custom-user-data-scripts/gradle-common-custom-user-data.gradle
+++ b/common-custom-user-data-scripts/gradle-common-custom-user-data.gradle
@@ -39,7 +39,7 @@ void tagIde() {
     if (project.hasProperty('android.injected.invoked.from.ide')) {
         buildScan.tag 'Android Studio'
         if (project.hasProperty('android.injected.studio.version')) {
-            buildScan.value 'AS Version', project.getProperty('android.injected.studio.version')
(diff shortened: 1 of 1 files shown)


common-custom-user-data-scripts/gradle-common-custom-user-data.gradle
```diff
@@ -38,6 +38,9 @@ void tagOs() {
 void tagIde() {
     if (project.hasProperty('android.injected.invoked.from.ide')) {
         buildScan.tag 'Android Studio'
+        if (project.hasProperty('android.injected.studio.version')) {
+            buildScan.value 'AS Version', project.getProperty('android.injected.studio.version')
+        }
```
etiennestuder, line 43: Given the existing tag is called `Android Studio`, I would call the custom value `Android Studio version`. Or we also rename the tag to `AS`. WDYT?

runningcode: That's a good point. The longer name isn't a problem anymore either since we can now edit tags. Updated.

etiennestuder:
@runningcode LGTM. Please merge and close.

PR #35, opened 2020-08-28
Add support for tagging old Idea releases

britter:
Releases prior to IntelliJ 2020.2 used to add a system property of the
form `-Didea.version2019.4` to the command line. So instead of having a
property with key `idea.version` and the version as value there was a
key that carried the version information.

### DIFF
--- a/common-custom-user-data-scripts/maven-common-custom-user-data.groovy
+++ b/common-custom-user-data-scripts/maven-common-custom-user-data.groovy
@@ -37,7 +37,7 @@ static void tagOs(def buildScan) {
 }
 
 static void tagIde(def buildScan) {
-    if (System.getProperty('idea.version')) {
+    if (System.getProperty('idea.version') || sysPropertyKeyStartingWith("idea.version")) {
         buildScan.tag 'IntelliJ IDEA'
     } else if (System.getProperty('eclipse.buildId')) {
         buildScan.tag 'Eclipse'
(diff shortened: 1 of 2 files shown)


common-custom-user-data-scripts/maven-common-custom-user-data.groovy
```diff
@@ -37,7 +37,7 @@ static void tagOs(def buildScan) {
 }
```
etiennestuder, line 38: I will soon delete the Groovy file for Maven. But, can you please also update the custom data script for Gradle? Thanks.

etiennestuder (changes requested):

guylabs (approved):

common-custom-user-data-scripts/maven-common-custom-user-data.groovy
britter: Gradle is not affected by this issue since the system property is set correctly in the from `-Didea.version=<version>` for Gradle invocations. At least for IntelliJ 2020.1 and 2020.2. I haven't checked older IntelliJ releases.

etiennestuder: My bad. Sorry about that.

etiennestuder (approved):

snoopcheri (approved):

facewindu (approved):

PR #37, opened 2020-09-04
Inject the BuildCacheApi in the custom groovy script

facewindu:
This PR accesses the BuildCache API and injects it in the default groovy script, to be used for custom configuration of the build cache feature of Gradle Enterprise.

This introduces a backward incompatibility, as we now compile against a 1.7-ish gradle enterprise maven extension version.
Do we want move things around to keep backward compatibility ? Keep in mind also that the groovy script must be evalluated only once, so we have to add logic to call with the proper injected parametrs (buildScan or buildScan/buildCache) depending on the version of the provided Maven extension version, at runtime.


common-custom-user-data-maven-extension/pom.xml
```diff
@@ -65,7 +65,7 @@
         <dependency>
             <groupId>com.gradle</groupId>
             <artifactId>gradle-enterprise-maven-extension</artifactId>
-            <version>1.6.4</version>
+            <version>1.7-branch-francois_6769_maven_build_cache_api-prerelease</version>
```
facewindu, line 68: This is what I used locally, and will obviously have to change before release

common-custom-user-data-maven-extension/src/main/java/com/gradle/GroovyScriptUserData.java
```diff
@@ -28,20 +34,29 @@ private static File getScriptFile(MavenSession session) {
...
-    private static Binding prepareBinding(MavenSession session, BuildScanApi buildScan, Logger logger) {
+    private static Binding prepareBinding(MavenSession session, @Nullable BuildScanApi buildScan, @Nullable BuildCacheApi buildCache, Logger logger) {
         Binding binding = new Binding();
         binding.setVariable("project", session.getTopLevelProject());
         binding.setVariable("session", session);
-        binding.setVariable("buildScan", buildScan);
+        if (buildScan != null) {
+            binding.setVariable("buildScan", buildScan);
+        } else {
+            logger.debug("Cannot bind buildScan API.");
```
britter, line 53: I wonder whether this should be a warning to make users aware they are using an outdated version of the Gradle Enterprise Maven Extension.

common-custom-user-data-maven-extension/src/main/java/com/gradle/CommonCustomUserDataMavenExtension.java
```diff
@@ -29,9 +30,10 @@ public void afterProjectsRead(MavenSession session) throws MavenExecutionExcepti
         if (buildScan != null) {
             logger.debug("Capturing custom user data in build scan");
             CustomUserData.addToBuildScan(buildScan);
-            GroovyScriptUserData.addToBuildScan(session, buildScan, logger);
             logger.debug("Finished capturing custom user data in build scans");
         }
+        BuildCacheApi buildCache = BuildCacheApiAccessor.lookup(container, getClass());
```
britter, line 35: I think it's possible to make this backwards compatible by typing this with `Object`.

britter (approved):
Don't forget to add some tests to our internal test suite.

common-custom-user-data-maven-extension/src/main/java/com/gradle/CommonCustomUserDataMavenExtension.java
marcphilipp: I think it's ok to break backward compatibility here since this code also serves as a showcase for programmatic configuration.

❌ We should increment the major version number, though.

common-custom-user-data-maven-extension/src/main/java/com/gradle/BuildCacheApiAccessor.java
```diff
@@ -0,0 +1,61 @@
...
+import org.apache.maven.MavenExecutionException;
+import org.codehaus.plexus.PlexusContainer;
+import org.codehaus.plexus.classworlds.realm.ClassRealm;
+import org.codehaus.plexus.component.repository.exception.ComponentLookupException;
+
+import java.util.Optional;
+
+import static java.util.Comparator.comparing;
+
+final class BuildCacheApiAccessor {
```
marcphilipp, line 13: This is largely copied fro `BuildScanApiAccessor`, right? Can we replace it with a single static parameterized method `<T> T lookup(Class<T> componentClass, PlexusContainer container, Class<?> extensionClass)` in a `ComponentAccessor` class?

common-custom-user-data-maven-extension/src/main/java/com/gradle/GroovyScriptUserData.java
marcphilipp: You mean below, right? Do we get into these `else` blocks at all? If so, in which cases?

marcphilipp (changes requested):

common-custom-user-data-maven-extension/src/main/java/com/gradle/GroovyScriptUserData.java
guylabs: When the `BuildCacheApi` component couldn't be found (https://github.com/gradle/gradle-enterprise-build-config-samples/pull/37/files#diff-252fd27e9805a7ae34557c0a8426795eR48) This can happen when a user uses an old Maven extension version that doesn't have the `BuildCacheApi` class I guess.

common-custom-user-data-maven-extension/pom.xml
guylabs: Used `1.6.5` for the upcoming release.

common-custom-user-data-maven-extension/src/main/java/com/gradle/BuildCacheApiAccessor.java
guylabs: Done.

common-custom-user-data-maven-extension/src/main/java/com/gradle/CommonCustomUserDataMavenExtension.java
guylabs: Increased it to `1.3-SNAPSHOT`

common-custom-user-data-maven-extension/src/main/java/com/gradle/GroovyScriptUserData.java
marcphilipp: But then the we'd get a `ClassNotFoundError` for `BuildCacheApi`, wouldn't we?

common-custom-user-data-maven-extension/src/main/java/com/gradle/ApiAccessor.java
```diff
@@ -43,19 +39,18 @@ private static void ensureBuildScanApiIsAccessible(Class<?> extensionClass) thro
         }
     }
 
-    private static BuildScanApi lookupBuildScanApi(PlexusContainer container) throws MavenExecutionException {
-        if (!container.hasComponent(BUILD_SCAN_API_CONTAINER_OBJECT)) {
+    private static <T> T lookupClass(Class<T> componentClass, String component, PlexusContainer container) throws MavenExecutionException {
+        if (!container.hasComponent(component)) {
```
marcphilipp, line 43: ❓ Can we use `PlexusContainer.hasComponent(Class)` and `PlexusContainer.lookup(Class)` here? That would allow us to get rid of the `component` String.

common-custom-user-data-maven-extension/src/main/java/com/gradle/ApiAccessor.java
```diff
@@ -10,31 +9,28 @@
...
-final class BuildScanApiAccessor {
+final class ApiAccessor {
 
-    private static final String PACKAGE = "com.gradle.maven.extension.api.scan";
-    private static final String BUILD_SCAN_API_CONTAINER_OBJECT = PACKAGE + ".BuildScanApi";
-
-    static BuildScanApi lookup(PlexusContainer container, Class<?> extensionClass) throws MavenExecutionException {
-        ensureBuildScanApiIsAccessible(extensionClass);
-        return lookupBuildScanApi(container);
+    static <T> T lookup(Class<T> componentClass, String componentPackage, String componentRole, PlexusContainer container, Class<?> extensionClass) throws MavenExecutionException {
```
marcphilipp, line 14: ❌ `componentPackage` and `componentRole` should be derived from `componentClass`. `componentRole` is just the fully-qualified class name, isn't it?

marcphilipp (changes requested):

common-custom-user-data-maven-extension/src/main/java/com/gradle/ApiAccessor.java
guylabs: Done

common-custom-user-data-maven-extension/src/main/java/com/gradle/ApiAccessor.java
guylabs: Done

common-custom-user-data-maven-extension/src/main/java/com/gradle/GroovyScriptUserData.java
guylabs: Yes I think so as well. Should we remove these logs then?

common-custom-user-data-maven-extension/src/main/java/com/gradle/ApiAccessor.java
```diff
@@ -43,19 +39,18 @@ private static void ensureBuildScanApiIsAccessible(Class<?> extensionClass) thro
...
 
-    private static BuildScanApi lookupBuildScanApi(PlexusContainer container) throws MavenExecutionException {
-        if (!container.hasComponent(BUILD_SCAN_API_CONTAINER_OBJECT)) {
+    private static <T> T lookupClass(Class<T> componentClass, PlexusContainer container) throws MavenExecutionException {
+        if (!container.hasComponent(componentClass)) {
             return null;
         } else {
             try {
-                return (BuildScanApi) container.lookup(BUILD_SCAN_API_CONTAINER_OBJECT);
+                return componentClass.cast(container.lookup(componentClass));
```
marcphilipp, line 47: This should be enough, right?

```suggestion
                return container.lookup(componentClass);
```

common-custom-user-data-maven-extension/src/main/java/com/gradle/CommonCustomUserDataMavenExtension.java
```diff
@@ -25,13 +26,14 @@ public CommonCustomUserDataMavenExtension(PlexusContainer container, Logger logg
...
-        BuildScanApi buildScan = BuildScanApiAccessor.lookup(container, getClass());
+        BuildScanApi buildScan = ApiAccessor.lookup(BuildScanApi.class, container, getClass());
         if (buildScan != null) {
             logger.debug("Capturing custom user data in build scan");
             CustomUserData.addToBuildScan(buildScan);
-            GroovyScriptUserData.addToBuildScan(session, buildScan, logger);
             logger.debug("Finished capturing custom user data in build scans");
         }
+        BuildCacheApi buildCache = ApiAccessor.lookup(BuildCacheApi.class, container, getClass());
+        GroovyScriptUserData.addToApis(session, buildScan, buildCache, logger);
```
etiennestuder, line 36: Let's do like for build scans and have a class called `CustomBuildCacheConfig` with `CustomBuildCacheConfig.configureBuildCache(BuildCacheApi buildCache). The implementation can be empty for now.

marcphilipp (changes requested):

common-custom-user-data-maven-extension/src/main/java/com/gradle/GroovyScriptUserData.java
marcphilipp: I think so.

etiennestuder (changes requested):

common-custom-user-data-maven-extension/src/main/java/com/gradle/GroovyScriptUserData.java
guylabs: Removed

common-custom-user-data-maven-extension/src/main/java/com/gradle/CommonCustomUserDataMavenExtension.java
guylabs: Done

common-custom-user-data-maven-extension/src/main/java/com/gradle/ApiAccessor.java
guylabs: Done

guylabs: Done.

common-custom-user-data-maven-extension/src/main/java/com/gradle/CustomBuildCacheConfig.java
```diff
@@ -0,0 +1,13 @@
+package com.gradle;
+
+import com.gradle.maven.extension.api.cache.BuildCacheApi;
+
+final class CustomBuildCacheConfig {
+
+    static void configureBuildCache(BuildCacheApi buildCache) {
+    }
```
marcphilipp, line 8: ⭕ Maybe add a `// custom config goes here` comment here?

marcphilipp (approved):

common-custom-user-data-maven-extension/src/main/java/com/gradle/ApiAccessor.java
```diff
@@ -10,31 +9,28 @@
 
 import static java.util.Comparator.comparing;
 
-final class BuildScanApiAccessor {
+final class ApiAccessor {
```
britter, line 12: This class is referenced in the Maven extension documentation in the programmatic configuration section. We need to adjust that section.

britter (approved):
👍

common-custom-user-data-maven-extension/src/main/java/com/gradle/ApiAccessor.java
guylabs: Yep nice catch. Marc already mentioned this on the other pull request and there is an after release step activity to update it.

common-custom-user-data-maven-extension/src/main/java/com/gradle/ApiAccessor.java
marcphilipp: This could use the class now again.

guylabs: yep fixed.

common-custom-user-data-maven-extension/src/main/java/com/gradle/GroovyScriptUserData.java
etiennestuder: @marcphilipp @guylabs Since you both 'think so' but do not know for sure, we need to ensure what the actual behavior is. @guylabs can you please check it? Thanks.

guylabs: It will never evaluate the Groovy script if one of the API's are `null`. See https://github.com/gradle/gradle-enterprise-build-config-samples/blob/79cdd2aeb9f988f7d653196d6493033363c20d22/common-custom-user-data-maven-extension/src/main/java/com/gradle/GroovyScriptUserData.java#L18-L30 So all good.

etiennestuder: No, that is not waht the code is saying.

PR #49, opened 2021-01-13
Convert common-custom-user-data.gradle script to a Gradle plugin

bigdaz:
Adds a Gradle plugin implementation that should replace the existing common-custer-user-data script.

- Applies all of the tags/links/values that are applied by the parallel Maven Extension
   - The Gradle Plugin implementation attempts to diverge as little as possible from the parallel Maven Extension code
   - In a couple of cases, the Maven Extension was modified to match the Gradle Plugin implemetation
- Detects Android Studio and tags the build scan appropriately
- Collects `maxParallelForks` and configured system properties for all `Test` tasks
- External dependencies (e.g. Guava) have been avoided in order to avoid polluting the build script classpath

Still to do:
- Support for publishing to `plugins.gradle.org`
- Reset the `CHANGELOG.md` as part of initial release


common-custom-user-data-gradle-plugin/src/main/java/com/gradle/CustomBuildScanConfig.java
```diff
@@ -0,0 +1,387 @@
...
+
+    @SuppressWarnings("UnstableApiUsage")
+    private static String execAndGetStdOut(String... args) {
+        Runtime runtime = Runtime.getRuntime();
+        Process process;
+        try {
+            process = runtime.exec(args);
+        } catch (IOException e) {
+            throw new RuntimeException(e);
+        }
```
runningcode, line 272:
```suggestion
        } finally {
            process.destroyForcibly();
        }
```

common-custom-user-data-gradle-plugin/src/main/java/com/gradle/CustomBuildScanConfig.java
```diff
@@ -0,0 +1,387 @@
...
+
+final class CustomBuildScanConfig {
+
+    static void configureBuildScan(BuildScanExtension buildScan, Gradle gradle) {
+        tagOs(buildScan);
+        tagIde(buildScan, gradle);
+        tagCiOrLocal(buildScan);
+        addCiMetadata(buildScan);
+        addGitMetadata(buildScan);
+        captureTestProperties(buildScan, gradle);
```
runningcode, line 33: We should have a way of disabling `captureTestProperties` it can be spammy in some builds.

common-custom-user-data-gradle-plugin/src/main/java/com/gradle/CustomBuildScanConfig.java
bigdaz: I don't think this will work, since the process is used in the subsequent try/finally block. But you're right, we could be more careful about cleaning up the process.

common-custom-user-data-gradle-plugin/CHANGELOG.md
```diff
@@ -0,0 +1,10 @@
+# Changelog
+All notable changes to this project will be documented in this file.
+
+## [Unreleased]
+
+## [1.0] - 2021-01-11
+Initial release.
+
+[Unreleased]: https://github.com/olivierlacan/keep-a-changelog/compare/common-custom-user-data-gradle-plugin-1.0...HEAD
```
etiennestuder, line 9: I don't understand this link.

common-custom-user-data-gradle-plugin/build.gradle
```diff
@@ -0,0 +1,24 @@
...
+
+group = "com.gradle"
+version = "0.1"
+
+repositories {
+    gradlePluginPortal()
+}
+
+dependencies {
+    compileOnly "com.gradle:gradle-enterprise-gradle-plugin:3.5.1"
```
etiennestuder, line 14: Please mention to @jonnybbb that the GE plugin release process needs to be adjusted to also bump up the version here.

common-custom-user-data-gradle-plugin/settings.gradle
```diff
@@ -0,0 +1 @@
+rootProject.name = 'common-custom-user-data-gradle-plugin'
```
etiennestuder, line 1: I suggest to also enable (local) build caching via gradle.properties.

common-custom-user-data-gradle-plugin/src/main/java/com/gradle/CustomBuildScanConfig.java
etiennestuder: I suggest to wait with adding an extension until we have a first version ready to ship.

etiennestuder (changes requested):
LGTM. Only a few minor comments.

Fyi, I did not revisit in depth the refactoring of the former duplication in how we capture CI custom values and links. I assume Daz tested that these are still working.

common-custom-user-data-gradle-plugin/src/main/java/com/gradle/CustomBuildScanConfig.java
bigdaz: Yeah, I implemented this in a branch already, but using a simple system property instead of an extension. This will make it easier to enable/disable any optional build scan features without changing the actual build. These changes can be made persistent via `gradle.properties`.

I will submit this as a separate PR, which also includes functionality to optionally capture the output of `ps` and `jps` as custom values.

common-custom-user-data-gradle-plugin/CHANGELOG.md
bigdaz: I copied this from the Maven Extension: it doesn't work yet, but it should show changes since the last release.
In fact, it's likely the changelog will need to be updated as part of the initial release.

common-custom-user-data-gradle-plugin/build.gradle
bigdaz: 👍

common-custom-user-data-gradle-plugin/settings.gradle
bigdaz: Good point, will do.

PR #50, opened 2021-01-14
Add optional custom values to Gradle Build Scans

bigdaz:
This PR introduces 3 main changes:
1. Extracts all utility methods into a separate `Utils` class for both the Gradle Plugin and Maven Extension
2. Adds the option for `jps` and `ps` output to be added as custom values. These are enabled via system props.
3. Allows the `Test` task configuration to _not_ be added as custom values. This is disabled via system prop.


common-custom-user-data-gradle-plugin/src/main/java/com/gradle/CustomBuildScanConfigExtras.java
```diff
@@ -0,0 +1,77 @@
...
+import org.gradle.api.Task;
+import org.gradle.api.invocation.Gradle;
+import org.gradle.api.tasks.testing.Test;
+
+import static com.gradle.Utils.*;
+
+final class CustomBuildScanConfigExtras {
+
+    static void configureBuildScan(BuildScanExtension buildScan, Gradle gradle) {
+        if (isCaptureEnabled("TestTaskConfiguration", true)) {
```
etiennestuder, line 14: I would set it to false by default.

etiennestuder: (different of today's behavior, I know)

common-custom-user-data-gradle-plugin/src/main/java/com/gradle/Utils.java
```diff
@@ -0,0 +1,157 @@
...
+import java.io.UnsupportedEncodingException;
+import java.net.URLEncoder;
+import java.nio.charset.Charset;
+import java.nio.charset.StandardCharsets;
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
+import java.util.Properties;
+import java.util.concurrent.TimeUnit;
+
+class Utils {
```
etiennestuder, line 18: can be final

common-custom-user-data-gradle-plugin/src/main/java/com/gradle/CustomBuildScanConfigExtras.java
bigdaz: Sure. My intent was to preserve the existing functionality of the script.
Are you OK with Test task parallelism not being reported by default, too? (These are currently controlled by the same system property).

common-custom-user-data-maven-extension/src/main/java/com/gradle/Utils.java
```diff
@@ -0,0 +1,128 @@
...
+import java.io.InputStreamReader;
+import java.io.Reader;
+import java.io.UnsupportedEncodingException;
+import java.net.URLEncoder;
+import java.nio.charset.Charset;
+import java.nio.charset.StandardCharsets;
+import java.util.Properties;
+import java.util.concurrent.TimeUnit;
+
+class Utils {
```
etiennestuder, line 16: can be final

etiennestuder (changes requested):

common-custom-user-data-gradle-plugin/src/main/java/com/gradle/CustomBuildScanConfigExtras.java
etiennestuder: I think it is ok. we can still tell people in the trial navigator to set these properties to true initially, and to turn them off if it is too much noise, but by then we already have a bseline.

etiennestuder:
Nice refactoring. My feedback is only of minor nature.

common-custom-user-data-gradle-plugin/src/main/java/com/gradle/Utils.java
bigdaz: Done

common-custom-user-data-maven-extension/src/main/java/com/gradle/Utils.java
bigdaz: Done

common-custom-user-data-gradle-plugin/src/main/java/com/gradle/CustomBuildScanConfigExtras.java
bigdaz: Done

PR #56, opened 2021-01-28
Improve documentation for common-custom-user-data-maven-extension

bigdaz:


common-custom-user-data-maven-extension/.mvn/gradle-enterprise-custom-user-data.groovy
```diff
@@ -0,0 +1,15 @@
...
+log.debug('Evaluating my custom Groovy script')
+
+// Add a build scan tag with the project name
+buildScan.tag(project.name)
+
+// Add a custom value based on a property of the `Session`
+buildScan.value('parallel', session.parallel as String)
+
+// Enable storing in the remote build cache based on the presence of a `CI` environment variable
+buildCache.getRemote().setStoreEnabled(session.systemProperties.containsKey('CI'))
```
etiennestuder, line 15: I think all of the above could be written more Groovy-style (parenthesis, getters, setters).

common-custom-user-data-maven-extension/src/main/resources/gradle-enterprise.sample.xml
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0" encoding="UTF-8" standalone="yes" ?>
+
+<!-- Minimal gradle-enterprise.xml configuration file for Maven.
+
+     This file specifies only the URL of the Gradle Enterprise server, assuming that the rest of the Gradle Enterprise
+     configuration is provided programmatically in the extension.
+-->
```
etiennestuder, line 7: Please add a comment reminding people to rename this file in order for it to take affect.

common-custom-user-data-maven-extension/src/main/resources/gradle-enterprise.sample.xml
```diff
@@ -0,0 +1,16 @@
...
+
+<gradleEnterprise
+    xmlns="https://www.gradle.com/gradle-enterprise-maven" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:schemaLocation="https://www.gradle.com/gradle-enterprise-maven https://www.gradle.com/schema/gradle-enterprise-maven.xsd">
+  <server>
+    <url>https://enterprise-samples.gradle.com</url>
```
etiennestuder, line 13: I think we used https://my-gradle-enterprise-server.com in other places. Can you double check and align?

common-custom-user-data-maven-extension/README.md
```diff
@@ -30,7 +44,26 @@ This approach has a number of benefits:
...
 - Standardize the configuration for connecting to Gradle Enterprise and Gradle Build Cache in your organization, removing the need for each project to specify this configuration.
 
-If your customized extension provides all required Gradle Enterprise configuration, then a consumer project will get all the benefits of Gradle Enterprise simply by applying the extension. The plugin sources provide a placeholder and example code to get you started.
+The project sources provide a good template to get started with your own Maven extension. 
+Refer to the [Javadoc](https://docs.gradle.com/enterprise/maven-extension/api/) for more details on the key types available for use.
+
+#### Providing Gradle Enterprise configuration in your custom Maven extension
+
+Using a custom extension, it is possible to completely remove the need for each project to declare a `gradle-enterprise.xml` file.
+Much of the configuration can be supplied programmatically. See `CustomBuildScanConfig` and `CustomBuildCacheConfig` for examples.
```
etiennestuder, line 53: I suggest to link to these two classes.

common-custom-user-data-maven-extension/README.md
```diff
@@ -30,7 +44,26 @@ This approach has a number of benefits:
...
+
+At this time, the Gradle Enterprise server URL and credentials cannot be provided programmatically, and must be specified in a `gradle-enterprise.xml` file.
```
etiennestuder, line 55: Not sure why you mention the credentials here. They only need to be set on the remote build cache - which can be done [programmatically](https://docs.gradle.com/enterprise/maven-extension/api/com/gradle/maven/extension/api/cache/Credentials.html).

common-custom-user-data-maven-extension/README.md
```diff
@@ -30,7 +44,26 @@ This approach has a number of benefits:
...
+To avoid adding this file to every project, this configuration file can be added to the `src/main/resources` directory so that it will be discovered by Gradle Enterprise when loading the extension.
```
etiennestuder, line 56: I would make it a bit more explicit that putting the file into src/main/resources means it will be packaged in the custom extension and discovered when loading the Gradle Enterprise extension.

common-custom-user-data-maven-extension/README.md
```diff
@@ -30,7 +44,26 @@ This approach has a number of benefits:
...
+This project includes a `gradle-enterprise.sample.xml` file, which should be modified with your server URL and renamed to `gradle-enterprise.xml`.
+
+Once your custom extension provides all required Gradle Enterprise configuration, then a consumer project will get all the benefits of Gradle Enterprise simply by applying the extension.
+
+#### Running custom build scan configuration logic once
```
etiennestuder, line 61: I would omit this block. This API is not needed anymore. It was needed when we had the groovy plugin execute groovy scripts and it was done for each sub-project.

common-custom-user-data-maven-extension/src/main/java/com/gradle/CustomBuildScanEnhancements.java
```diff
@@ -0,0 +1,253 @@
...
+import java.util.Properties;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+import static com.gradle.Utils.*;
+
+/**
+ * Adds a standard set of useful tags, links and custom values to all build scans published.
+ */
+final class CustomBuildScanEnhancements {
```
runningcode, line 14: One thing we could do is put this file in a shared library so that the maven extension and the gradle plugin can access it and we don't have to duplicate it like the spring project does: https://github.com/spring-io/gradle-enterprise-conventions

I'm guess we might not though because that would make it too complicated for customers to use the custom maven extension?

etiennestuder: Yexp, that's the reason we live with the duplication.

common-custom-user-data-maven-extension/.mvn/gradle-enterprise-custom-user-data.groovy
bigdaz: For sure. (I just copied exactly what was in the user manual)

bigdaz: I only updated the last line to use property syntax instead of setters/getters.
I think the code is clearer with the parentheses left in.

common-custom-user-data-maven-extension/src/main/resources/gradle-enterprise.sample.xml
bigdaz: Actually we use https://your-gradle-enterprise-server.com. I'll make sure this is consistent.

common-custom-user-data-maven-extension/src/main/java/com/gradle/CustomBuildScanEnhancements.java
bigdaz: We would also have to adapt to `com.gradle.maven.extension.api.scan.BuildScanApi` vs `com.gradle.scan.plugin.BuildScanExtension`, which would add complexity that isn't desired in a example app.

PR #59, opened 2021-02-02
Supports Hudson CI environments

kyhule:
Uses the `HUDSON_URL` environment variable set by Hudson CI to determine
if build is being run by a CI instance and to capture CI metadata.

### DIFF
--- a/common-custom-user-data-gradle-plugin/src/main/java/com/gradle/CustomBuildScanEnhancements.java
+++ b/common-custom-user-data-gradle-plugin/src/main/java/com/gradle/CustomBuildScanEnhancements.java
@@ -64,7 +64,7 @@ final class CustomBuildScanEnhancements {
     }
 
     private static void captureCiMetadata(BuildScanExtension buildScan) {
-        if (isJenkins()) {
+        if (isJenkins() || isHudson()) {
             if (envVariablePresent("BUILD_URL")) {
                 buildScan.link("Jenkins build", envVariable("BUILD_URL"));
             }
@@ -183,7 +183,7 @@ final class CustomBuildScanEnhancements {
     }
 
     private static boolean isCi() {
-        return isGenericCI() || isJenkins() || isTeamCity() || isCircleCI() || isBamboo() || isGitHubActions() || isGitLab() || isTravis() || isBitrise();
+        return isGenericCI() || isJenkins() || isHudson() || isTeamCity() || isCircleCI() || isBamboo() || isGitHubActions() || isGitLab() || isTravis() || isBitrise();
     }
 
     private static boolean isGenericCI() {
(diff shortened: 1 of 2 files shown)


common-custom-user-data-gradle-plugin/src/main/java/com/gradle/CustomBuildScanEnhancements.java
```diff
@@ -64,7 +64,7 @@ private static void captureCiOrLocal(BuildScanExtension buildScan) {
     }
 
     private static void captureCiMetadata(BuildScanExtension buildScan) {
-        if (isJenkins()) {
+        if (isJenkins() || isHudson()) {
             if (envVariablePresent("BUILD_URL")) {
                 buildScan.link("Jenkins build", envVariable("BUILD_URL"));
```
etiennestuder, line 69: Shouldn't it say `Hudson build` in this new case?

common-custom-user-data-gradle-plugin/src/main/java/com/gradle/CustomBuildScanEnhancements.java
```diff
@@ -194,6 +194,10 @@ private static boolean isJenkins() {
         return envVariablePresent("JENKINS_URL");
     }
 
+    private static boolean isHudson() {
+        return envVariablePresent("HUDSON_URL");
```
etiennestuder, line 198: Something does not add up, I think. Here we check for HUDSON_URL but in captureCiMetadata we wetch BUILD_URL, even in the Hudson case.

etiennestuder (changes requested):

common-custom-user-data-gradle-plugin/src/main/java/com/gradle/CustomBuildScanEnhancements.java
kyhule: Yes, that would probably be preferable. I was hesitant to add an entire new if block since the both set the same environment variables that are captured here (Jenkins is a fork of Hudson after all) but that might not always be the case so I can make a separate condition for Hudson.

common-custom-user-data-gradle-plugin/src/main/java/com/gradle/CustomBuildScanEnhancements.java
kyhule: `HUDSON_URL` is the root homepage of the Hudson instance (similar to `JENKINS_URL` being the root homepage of the Jenkins instances where as `BUILD_URL` is the url to the specific build of the job that ran.

etiennestuder: Got it. My bad. Thanks for the explanation.

etiennestuder (approved):

common-custom-user-data-gradle-plugin/src/main/java/com/gradle/CustomBuildScanEnhancements.java
etiennestuder: I went with something in the middle. We can still add a separate section if Jenkins/Hudson diverge from each other.

etiennestuder:
Thanks @kyhule !

PR #61, opened 2021-02-05
Rename script to have `.sh` ending.

runningcode:
This will help when parsing for specific file endings.


etiennestuder (changes requested):
@runningcode Please also rename create-custom-maven-wrapper. Also, the example invocation needs to be adjusted.

runningcode:
I also updated the versions update process document with the updated url

PR #62, opened 2021-02-08
Fix TeamCity integration

bigdaz:
The TeamCity 2020.2.2 no system properties are added to the build invocation
making our integration non-functional. This fix instead relies on env vars
to integrate with TeamCity.


etiennestuder (approved):
While I feel somewhat uncomfortable that we make this change without knowing what changed on TeamCity's side, the change seems to be sound for at least the latest TC version, TC 2020.2.

runningcode (approved):

PR #76, opened 2021-03-03
Fix TeamCity integration. Again.

bigdaz:


common-custom-user-data-gradle-plugin/src/main/java/com/gradle/CustomBuildScanEnhancements.java
```diff
@@ -18,269 +20,302 @@
...
-                addCustomValueAndSearchLink(buildScan, "CI plan", envVariable("bamboo_planName"));
-            }
-            if (envVariablePresent("bamboo_buildPlanName")) {
-                addCustomValueAndSearchLink(buildScan, "CI build plan", envVariable("bamboo_buildPlanName"));
-            }
-            if (envVariablePresent("bamboo_agentId")) {
-                addCustomValueAndSearchLink(buildScan, "CI agent", envVariable("bamboo_agentId"));
-            }
+        private void captureIde() {
+            gradle.projectsEvaluated(g -> {
```
runningcode, line 57: oh, is this because project properties are not yet available when this is applied as a settings plugin?
maybe we should add a comment here so we know why it is there.

common-custom-user-data-gradle-plugin/src/main/java/com/gradle/CustomBuildScanEnhancements.java
```diff
@@ -18,269 +20,302 @@
...
+                    api.tag("Dirty");
+                    api.value("Git status", gitStatus);
+                }
+            });
+        }
 
-    private static boolean isGitInstalled() {
-        return execAndCheckSuccess("git", "--version");
-    }
+        private static boolean isGitInstalled() {
```
runningcode, line 289: is the indentation change here intentional?

runningcode (commented):
thanks for looking in to this. just a quick comment.

runningcode (approved):

common-custom-user-data-gradle-plugin/src/main/java/com/gradle/CustomBuildScanEnhancements.java
```diff
@@ -8,6 +8,7 @@
 import org.gradle.api.invocation.Gradle;
 import org.gradle.api.tasks.testing.Test;
 
+import java.util.Optional;
```
runningcode, line 11: I think this was introduced in java 8. Are we at all worried about java 6 or 7 compat?

runningcode (approved):

common-custom-user-data-gradle-plugin/src/main/java/com/gradle/Utils.java
```diff
@@ -39,12 +36,12 @@ static boolean sysPropertyKeyStartingWith(String keyPrefix) {
...
 
-    static String envVariable(String name) {
-        return System.getenv(name);
-    }
-
-    static boolean envVariablePresent(String name) {
-        return isNotEmpty(envVariable(name));
+    static Optional<String> envVariable(String name) {
+        String value = System.getenv(name);
+        if (isNotEmpty(value)) {
```
runningcode, line 41: can you use `Optional.ofNullable` here?

common-custom-user-data-gradle-plugin/src/main/java/com/gradle/CustomBuildScanEnhancements.java
bigdaz: Gradle 5 requires at least Java 8 to run, so I think this is ok.

common-custom-user-data-gradle-plugin/src/main/java/com/gradle/Utils.java
bigdaz: We want to check for empty strings as well, and `Optional.ofNullable` doesn't do this.

PR #78, opened 2021-03-04
Enable stricter plugin validation.

runningcode:


etiennestuder:
@runningcode I somehow missed this PR. Interesting change. Should we also recommend this in our trials? Let's discuss f2f.

PR #79, opened 2021-03-31
Add scripts to run Gradle Enterprise Trial experiments

jthurne:


experiments/capture-build-scan-info.gradle
```diff
@@ -0,0 +1,49 @@
...
+def isTopLevelBuild = gradle.getParent() == null
+
+if (isTopLevelBuild) {
+    def version = GradleVersion.current().baseVersion
+    def atLeastGradle5 = version >= GradleVersion.version("5.0")
+    def atLeastGradle6 = version >= GradleVersion.version("6.0")
+    
+    if (atLeastGradle6) {
+        settingsEvaluated {
+            it.extensions["gradleEnterprise"].with {
```
runningcode, line 24: i think you can just replace this like so?
```suggestion
            gradleEnterprise {
```

experiments/capture-build-scan-info.gradle
```diff
@@ -0,0 +1,49 @@
...
+                   buildScanPublished { scan ->
+                       logScan(file("scans.csv"), scan)
+                   }
+               }
+            }
+        }
+    } else if (atLeastGradle5) {
+        projectsEvaluated {
+            rootProject {
+               it.extensions["buildScan"].with {
```
runningcode, line 35:
```suggestion
               buildScan {
```

experiments/run-experiment-01.sh
```diff
@@ -0,0 +1,101 @@
...
+ execute_first_build
+ execute_second_build
+ open_build_scan_comparison
+
+ cd ${project_dir}
+ printf "\n\033[00;32mDONE\033[0m\n"
+}
+
+collect_gradle_task() {
+  read -p "What gradle task do you want to run? (build) " task
```
runningcode, line 30:
```suggestion
  read -p "What Gradle task do you want to run? (build) " task
```

experiments/README.md
```diff
@@ -0,0 +1,6 @@
+# Experiments Automation
+
+Scripts that automate many of the Gradle Enterprise trial experiments. Using
+these scripts will help you to avoid many of the common mistakes people make as
+they work through the various experiments. For example, the Experiment 1 script
+ensures that the `--no-build-cache` flag is passed on every build invokation.
```
runningcode, line 6: 👍

runningcode (approved):
Looks great!

experiments/capture-build-scan-info.gradle
jthurne: Weirdly, when I tried that, it didn't work. I get an error like:

```
Could not find method gradleEnterprise() for arguments [capture_build_scan_info_7nzf7873k91ckmu9lx8hsbm6z$_run_closure1$_closure3@25f76ec3] on build of type org.gradle.invocation.DefaultGradle.
```

I always have trouble getting my head around the way in which variables are scoped and such in an init script. There's so many nested closures that run at different times and are fed different contexts.....

experiments/capture-build-scan-info.gradle
jthurne: When I tried this one, it actually worked. Maybe because the closer is invoked within the context of the root project. However, I think I'm going to keep it the way it is just so that this block is more consistent with the Gradle 6 block.

experiments/capture-build-scan-info.gradle
bigdaz: Nit: I think a named closure delegate is more readable:
```
settingsEvaluated { settings ->
   settings.extensions["gradleEnterprise"]... 
```

experiments/run-experiment-01.sh
```diff
@@ -0,0 +1,100 @@
...
+  mkdir -p "${experiment_dir}"
+}
+
+clone_project() {
+   info "Creating a clean clone of the project."
+
+   local clone_dir="${experiment_dir}/${project_name}"
+
+   rm -rf "${clone_dir}"
+   git clone "${project_dir}" "${clone_dir}"
```
bigdaz, line 49: `git clone --depth=1` will reduce the file copying required here

experiments/run-experiment-01.sh
```diff
@@ -0,0 +1,100 @@
...
+  local scan_id=()
+  # This isn't the most robust way to read a CSV,
+  # but we control the CSV so we don't have to worry about various CSV edge cases
+  while IFS=, read -r field_1 field_2 field_3; do
+     base_url+=("$field_1")
+     scan_id+=("$field_2")
+     scan_url+=("$field_3")
+  done < scans.csv
+
+  read -p "Press enter to to open the build scan in your default browser."
```
bigdaz, line 77: I'm not sure why you added a prompt here

experiments/capture-build-scan-info.gradle
jthurne: I like this. It helps to clarify what object is actually available and in-scope.

experiments/capture-build-scan-info.gradle
```diff
@@ -0,0 +1,49 @@
...
+
+if (isTopLevelBuild) {
+    def version = GradleVersion.current().baseVersion
+    def atLeastGradle5 = version >= GradleVersion.version("5.0")
+    def atLeastGradle6 = version >= GradleVersion.version("6.0")
+    
+    if (atLeastGradle6) {
+        settingsEvaluated {
+            it.extensions["gradleEnterprise"].with {
+               buildScan {
```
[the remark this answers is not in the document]
jthurne: That's a good idea.

Or...would it be better to do that in a separate init script so that this one is singularly focused on capturing build scan data?  In the 2nd init script, we could make sure all of the important flags for the trial are set (like this one).

Then again, I wasn't sure how much we wanted to automate with regards to configuring their build for the trail. There's a lot to be learned by going through the motions. Maybe I should write an init script that fails the build if the things we expect to  be set aren't set yet?

experiments/run-experiment-01.sh
jthurne: I found it rather disconcerting to have the browser open up automatically without any warning whatsoever. In particular, if I took my attention away from the build, it was a bit of a shocking experience to suddenly have the browser pop up to the top unbidden.

experiments/run-experiment-01.sh
```diff
@@ -0,0 +1,100 @@
...
+    'WindowsNT') browse=start ;;
+    *) browse=xdg-open ;;
+  esac
+  $browse "${base_url[0]}/c/${scan_id[0]}/${scan_id[1]}/task-inputs"
+}
+
+invoke_gradle() {
+  # The gradle --init-script flag only accepts a relative directory path. ¯\_(ツ)_/¯
+  local script_dir_rel=$(realpath --relative-to="$( pwd )" "${script_dir}")
+  local cmd="./gradlew --init-script ${script_dir_rel}/capture-build-scan-info.gradle -Dscan.tag.exp1 -Dscan.tag.${run_id} $@"
```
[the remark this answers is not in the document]
jthurne: Either that, or we set `publishAlways()` in an init script.

experiments/run-experiment-01.sh
jthurne: Interestingly, when I tried to use `--depth=1`, git printed out this warning:

```
warning: --depth is ignored in local clones; use file:// instead.
```

I now recall that git already does some pretty nifty things to optimize disk usage when cloning a local repository. Since commits are immutable, the original git repo, and the clone share many of the same files. I have a vague memory that it's actually using hard links to do this, but don't quote me on that.

experiments/run-experiment-01.sh
```diff
@@ -84,17 +251,160 @@ open_build_scan_comparison() {
...
+    read -p "$1"
+    echo "${RESTORE}"
+  fi
+}
+
+print_introduction() {
+  if [ "$_arg_wizard" == "on" ]; then
+    cat <<EOF
+${CYAN}
+                              ;x0K0d,
```
runningcode, line 315: Nice elephant!

experiments/README.md
```diff
@@ -0,0 +1,6 @@
+# Experiments Automation
```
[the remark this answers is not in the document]
jthurne: I demoed the script on Thursday during the sync-call, and everyone agrees that expanding these to allow users to re-run the experiments is good. However, we also decided it is still okay if we call these experiments: users can run experiments more than once.

But it's still a good thing to keep in mind, as it does guide some of our design decisions.

experiments/run-experiment-01.sh
jthurne: After the first demo (last Thursday) we decided to clone the repo from the original remote, which once again provided value to using `--depth=1`. So this has been incorporated.

experiments/run-experiment-01.sh
jthurne: We decided after the demo last Thursday to no longer automatically open the build comparison in the browser at all. Instead, we provide a set of links to the user after the experiment has run, which the user can click to open.

experiments/run-experiment-01.sh
jthurne: :)

etiennestuder (changes requested):
Nice work, @jthurne. Initial round of feedback in.

experiments/capture-build-scan-info.gradle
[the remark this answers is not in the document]
etiennestuder: @jthurne I shared my thoughts on that via slack. Single script seems fine to me. Not failing on missing things but adding them on the fly. and Reporting non-ideal configuration and rendering it on the cmd line at the end of the experiment.

experiments/01-validate-incremental-build.sh
```diff
@@ -0,0 +1,289 @@
...
+
+$(info "./${SCRIPT_NAME}")
+
+Congrats! You have completed this experiment.
+EOF
+  print_in_box "${text}"
+}
+
+main
+
```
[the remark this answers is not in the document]
jthurne: ✅  (done)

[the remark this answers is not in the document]
jthurne: ✅  Done except for getting rid of `--no-interactive`. To accelerate development, I'm using a tool called [argbash](https://argbash.io/) to auto-generate the command-line argument parsing code, and there doesn't seem to be an option to NOT generate the negating flags.

But once we're mostly done (assuming we stay with bash), we can come back and clean this up (by manually editing the autogenerated file).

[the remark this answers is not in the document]
jthurne: When not running in interactive mode, or when the config has not been saved, tasks is now required and the script fails if it isn't provided.

Is it okay to continue to use "assemble" as a default when running in interactive mode?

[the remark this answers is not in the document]
jthurne: ✅  Done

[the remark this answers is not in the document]
jthurne: ✅  Done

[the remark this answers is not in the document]
jthurne: ✅  Done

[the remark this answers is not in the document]
jthurne: ✅  Done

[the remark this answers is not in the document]
jthurne: ✅  Done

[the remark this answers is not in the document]
jthurne: ✅  Done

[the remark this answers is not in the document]
jthurne: ✅  Done

Unfortunately, it turns out that it isn't possible to apply the Common Custom User Data plugin from an init script if the Gradle Enterprise plugin has already been applied to the project outside of the init script. This is due to the plugins being loaded in separate class loaders.

But I did change things so that the experiment doesn't fail. Instead, the script outputs a warning at the end of the experiment and points the user to the documentation for the plugin.

If the user passes `--enable-gradle-enterprise`, then we can and do apply the plugin (because both plugins are loaded with the same class loader).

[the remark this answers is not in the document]
jthurne: ✅  Done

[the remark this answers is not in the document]
jthurne: The usage is currently autogenerated by argbash. Once we're done changing arguments and such, we can hand-edit the generated code so that the usage is more to our liking. I'll also look to see if there's a way to customize how argbash generates the usage.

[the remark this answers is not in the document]
jthurne: ✅  Done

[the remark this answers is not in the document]
jthurne: ✅ Done.

I'm not sure I got all of the quote handling quite right, but it's "mostly" right. Testing will help identify any additional issues with it.  Handling quotes in bash scripts is notoriously difficult.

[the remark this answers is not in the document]
jthurne: You can use `-e` or `--extra` to pass additional arguments to Gradle/Maven. I'm not totally sold on the argument name. Suggestions welcome!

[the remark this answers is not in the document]
jthurne: ✅ Done

It turns out to be very difficult to set captureTaskInputFiles in an init script, but there is a system property that one can set, so that's what I ended up doing.

Specifically, within an init script, if you try to set captureTaskFiles after the settings/project has been evaluated, you get an error. If you try to set it earlier, then any value the user sets it to in their settings or build script overwrite what the init script sets, basically nullifying the change.

But if you set it using the system property, then the system property overrides anything the user may set.

experiments/lib/verify-ge-configured.gradle
```diff
@@ -0,0 +1,51 @@
...
+            assert buildScan.server != null : "A Gradle Enterprise server to publish build scans to has not been configured. Please configure the build to publish build scans to a Gradle Enterprise instance and then try running the experiment again"
+            assert buildScan.captureTaskInputFiles : "Capturing task inputs should be enabled"
+            // TODO verify publishAlways has been enabled
+        }
+    }
+}
+
+def assertPluginExists(pluginManager, pluginName, pluginId) {
+    assert pluginManager.hasPlugin(pluginId) : "The ${pluginName} plugin is missing from this project. Please add '${pluginId}' to your build and then try running the experiment again"
+    // TODO Validate the plugin versions are the latest versions
```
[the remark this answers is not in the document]
jthurne: ✅ Done per our discussion earlier this week.

experiments/01-validate-incremental-build.sh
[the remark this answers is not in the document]
jthurne: ✅ Done

jthurne:
I've incorporated most of  @etiennestuder's feedback. To summarize all of the changes:

  - Added a script for the first Maven experiment
  - Moved the gradle scripts under a gradle/ subdirectory, and the maven script under a maven/ subdirectory.
  - The init scripts only terminate a build if the Gradle Enterprise plugin is missing.
  - If the Common Custom User Data plugin is missing, a warning is generated, but the experiment still executes.
  - Input is only prompted for when running in interactive mode. If not running in interactive mode, then an error is printed for missing required arguments (unless the script has already run once, in which case it loads the argument from the config file).
   - Renamed --wizard to --interactive
   - Renamed --task to --tasks
   - Added an --server argument to allow the user to set/override the GE server build scans are published to.
   - Added an --extra argument to allow the user to pass extra arguments to Gradle/Maven
   - Added an --enable-gradle-enterprise flag, that when set, will cause an init script to be added that adds the Gradle Enterprise and the Common Custom User Data plugins (only works on projects that don't already have the plugins applied).
   - Changed the Run IDs to be shorter (based on an encoded timestamp instead of being a UUID)
   - Various cosmetic changes
   
  
  Feedback that has not yet been incorporated:
  
   - Updating the introductory text that is first displayed when running with --interactive
   - Ensuring that all descriptive text ends with a period (.)
   - Reworking and reformatting the usage text (what prints with --help). This is hard because I'm currently using [argbash](https://argbash.io/) which autogenerates this code.

experiments/gradle/01-validate-incremental-build.sh
```diff
@@ -0,0 +1,291 @@
+#!/usr/bin/env bash
```
etiennestuder, line 1: When running without extra args, the following empty quotes are printed on the cmd line (they should not):

```
Running first build:
./gradlew '' --no-build-cache clean build -Dscan.tag.exp1-gradle -Dscan.tag.607d3f63
```

etiennestuder (changes requested):

jthurne:
> Thought this was really cool since I run these experiments manually all the time. Just checked out your branch to give it a try and I am seeing the following error when running with -i after starting the first build.

Thanks for trying it out @kyhule ! Just a warning that this is still very much an (active) work in progress.

From the error you posted, it looks like your local environment is missing the "realpath" command, which I had assumed was generally available on most systems. It looks like that assumption was faulty! I'll see if I can find an alternative approach.

jthurne:
@kyhule I just pushed up a fix that should resolve the issue you ran into.

kyhule:
> Thanks for trying it out @kyhule ! Just a warning that this is still very much an (active) work in progress.
Thanks for the heads up. I figured as such but was about to dig up my notes to run the experiments and remembered seeing this PR come through my inbox so figured I would check it out. Been meaning to write a way more rudimentary script for this so glad that I can take that off my list of things I'll never have time for.

kyhule:
@jthurne Gave it a go with your fix and finished as expected. Really digging the detailed output with the quick lints to comparisons. This is going to be so nice to have. Great work!

Looks as though `02-validate-build-caching-local-in-place.sh` is still using `realpath` 😬

jthurne:
> Looks as though 02-validate-build-caching-local-in-place.sh is still using realpath 😬

Whoops! I'll get that fixed up. Thanks for reporting it.

jthurne:
> > Looks as though 02-validate-build-caching-local-in-place.sh is still using realpath 😬
> 
> Whoops! I'll get that fixed up. Thanks for reporting it.

@kyhule all fixed!

build-validation/scripts/gradle/02-validate-build-caching-local-in-place.sh
```diff
@@ -166,63 +164,40 @@ print_introduction() {
...
-any) tasks should actually execute on the seond build (all of the task
-output should be used from the local cache).
+In this experiment, you will validate how well a given project leverages
+Gradle’s local build cache. A build is considered fully local-cache enabled if
+all tasks avoid performing any work because:
 
-The Gradle Solutions engineer will then work with you to figure out why some
-(if any) tasks ran on the second build, and how to optimize them to take
-advantage of the build cache.
+  * The tasks' inputs have not changed since their last invocation and
```
runningcode, line 171: This wording might suggest that the build cache can only use an output from the last build. This is the case for `up-to-date` checking in the first experiment but the build cache can re-use any output with the same inputs.

Maybe it might be better to say
`the tasks' inputs match the inputs of a previous invocation of the task - in this case the previous invocation in this experiment`

build-validation/scripts/gradle/02-validate-build-caching-local-in-place.sh
```diff
@@ -255,82 +227,40 @@ explain_second_build() {
...
-$(print_quick_links)
-
-Use the above links help you get started in your analysis.
-
-The first link is to a comparison of the two build scans. Comparisons show you
-what was different between two different build executions.
+The 'Summary' section below captures the configuration of the experiment and the
+two build scans that were published as part of running the experiment.  The
+build scan of the second build is particularly interesting since this is where
+you can inspect what tasks were not leveraging Gradle’s local build cache.
```
runningcode, line 246:
```suggestion
you can inspect which tasks were not leveraging Gradle’s local build cache.
```

build-validation/scripts/gradle/02-validate-build-caching-local-in-place.sh
```diff
@@ -255,82 +227,40 @@ explain_second_build() {
...
 
-The "Task execution summary" shows overall statistics for the execution of
-the second build. You can use this link to get a quick overview of where
-there may be overall opportunities to optimize.
+The 'Investigation Quick Links' section below allows quick navigation to the
+most relevant views in build scans to investigate what task outputs were fetched
```
runningcode, line 249:
```suggestion
most relevant views in build scans to investigate which task's outputs were fetched
```

build-validation/scripts/gradle/02-validate-build-caching-local-in-place.sh
```diff
@@ -255,82 +227,40 @@ explain_second_build() {
...
+from the cache and what tasks executed in the second build with cache misses,
```
runningcode, line 250:
```suggestion
from the cache and which tasks were executed in the second build because of cache misses,
```

build-validation/scripts/gradle/03-validate-build-caching-local-diff-place.sh
```diff
@@ -174,67 +172,42 @@ print_introduction() {
...
-output should be used from the local cache).
+In this experiment, you will validate how well a given project leverages
+Gradle’s local build cache when built from different locations. A build is
+considered fully local-cache enabled if all tasks avoid performing any work
+because:
 
-The Gradle Solutions engineer will then work with you to figure out why some
-(if any) tasks ran on the second build, and how to optimize them to take
-advantage of the build cache.
+  * The tasks' inputs have not changed since their last invocation and
```
runningcode, line 180: same comment as above

build-validation/scripts/gradle/03-validate-build-caching-local-diff-place.sh
```diff
@@ -245,13 +218,15 @@ explain_first_clone_project(){
...
 ${HEADER_COLOR}Check out project from Git for first build${RESTORE}
 
-For the first build, we are going create a fresh clone of the project.  That
-way, the experiment will be influenced by as few outside factors as
-possible.
+All configuration to run the experiment has been collected. For the first build,
+the Git repository that contains the project to validate will be checked out.
 
-We'll use a different clone of the project for the second build.
+A different check out of the project will be used for the second build.
```
runningcode, line 224: hmm, not sure if we need a `-` here
```suggestion
A different check-out of the project will be used for the second build.
```

build-validation/scripts/gradle/03-validate-build-caching-local-diff-place.sh
```diff
@@ -263,17 +238,11 @@ explain_first_build() {
...
-We are now ready to run the first build.
-
-We will execute 'clean ${tasks}'.
-
-We are invoking clean even though we just created a fresh clone because
-sometimes the clean task changes the order other tasks run in, which can
-impact how the build cache is utilized.
+Now that the project has been checked out, the first build can be run with the
+given Gradle tasks. The build will be invoked with the 'clean' task included and
+build caching disabled.
```
runningcode, line 243: is this true? we need to push entries in to the cache, no?

build-validation/scripts/gradle/03-validate-build-caching-local-diff-place.sh
```diff
@@ -303,84 +270,40 @@ explain_second_build() {
...
+two build scans that were published as part of running the experiment.  The
+build scan of the second build is particularly interesting since this is where
+you can inspect what tasks were not leveraging Gradle’s local build cache.
 
-The "Task execution summary" shows overall statistics for the execution of
-the second build. You can use this link to get a quick overview of where
-there may be overall opportunities to optimize.
+The 'Investigation Quick Links' section below allows quick navigation to the
+most relevant views in build scans to investigate what task outputs were fetched
+from the cache and what tasks executed in the second build with cache misses,
```
runningcode, line 293:
```suggestion
from the cache and which tasks were executed in the second build with cache misses,
```

build-validation/scripts/gradle/03-validate-build-caching-local-diff-place.sh
```diff
@@ -303,84 +270,40 @@ explain_second_build() {
...
+The 'Summary' section below captures the configuration of the experiment and the
+two build scans that were published as part of running the experiment.  The
+build scan of the second build is particularly interesting since this is where
+you can inspect what tasks were not leveraging Gradle’s local build cache.
 
-The "Task execution summary" shows overall statistics for the execution of
-the second build. You can use this link to get a quick overview of where
-there may be overall opportunities to optimize.
+The 'Investigation Quick Links' section below allows quick navigation to the
+most relevant views in build scans to investigate what task outputs were fetched
```
runningcode, line 292:
```suggestion
most relevant views in build scans to investigate which task outputs were fetched
```

build-validation/scripts/maven/01-validate-build-caching-local-in-place.sh
```diff
@@ -154,62 +153,40 @@ print_introduction() {
...
-any) goals should actually execute on the seond build (all of the goal
-output should be used from the local cache).
+In this experiment, you will validate how well a given Maven project leverages
+Gradle Enterprise’s local build cache. A build is considered fully local-cache
+enabled if all goals avoid performing any work because:
 
-The Gradle Solutions engineer will then work with you to figure out why some
-(if any) goals ran on the second build, and how to optimize them to take
-advantage of the build cache.
+  * The goals' inputs have not changed since their last invocation and
```
runningcode, line 160: same comment as before

runningcode (commented):
just some nits on the text! otherwise looks good

PR #80, opened 2021-04-01
Configure shellcheck to run on any push events

jthurne:
I'm not sure if this exactly what we want, but we can refine it after we see how
this behaves.

### DIFF
--- a/.github/workflows/shellcheck.yaml
+++ b/.github/workflows/shellcheck.yaml
@@ -14,7 +14,7 @@ jobs:
       env:
         GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
       with:
-        sh_checker_comment: false
+        sh_checker_comment: true
         sh_checker_exclude: "gradlew mvnw"
         sh_checker_shfmt_disable: true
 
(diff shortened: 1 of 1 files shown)


.github/workflows/shellcheck.yaml
```diff
@@ -0,0 +1,15 @@
...
+jobs:
+  shellcheck:
+    name: Shellcheck
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - name: Run ShellCheck
+      uses: ludeeus/action-shellcheck@master
+      with:
+        ignore: "./*/gradlew"
```
runningcode, line 15: maybe this should be 
```suggestion
        ignore: "**/gradlew"
```
Looks like it is still failing the build.

.github/workflows/shellcheck.yaml
```diff
@@ -0,0 +1,20 @@
+name: 'Shellcheck'
+
+on:
+  push
```
jthurne, line 4: I'm not sure we want shellcheck to run on every push, but I figured this would be a good place to start. If it's too aggressive or runs too frequently, we can scale this back to one of the more targeted events.

etiennestuder: +1 to your proposed strategy.

runningcode (approved):

PR #85, opened 2021-04-06
Tweak how shellcheck is run

jthurne:
If we trigger shellcheck on a pull-request event, then whatever it finds will be added as a comment to the pull-request, which is pretty nice.

This causes the check to run twice (once on the push event, and once on the pull-request event), but there isn't any significant downside to that.

### DIFF
--- a/.github/workflows/shellcheck.yaml
+++ b/.github/workflows/shellcheck.yaml
@@ -1,7 +1,9 @@
 name: 'Shellcheck'
 
 on:
-  push
+  push:
+  pull_request:
+    types: [synchronize]
 
 jobs:
   shellcheck:


etiennestuder:
> But ideally we would only run one check on the pull-request,

Why?

jthurne:
>> But ideally we would only run one check on the pull-request,

> Why?

A good question! Mostly so that the checks complete sooner (if there's a 2nd check, there's the risk that it will run longer than the other duplicate check). But maybe that risk is so low it doesn't matter.

It's a little wasteful of resources, but these are GitHub's resources, and as far as I know, we aren't paying for them. So there isn't a cost reason to limit the checks to just one.

jthurne:
I went ahead and removed the commit that tweaked when the push event is run, as I can't really think of a good reason try to only run the check once.

etiennestuder (approved):

runningcode:
For large projects with many checks where we are paying for the CPU cost of the github actions it would make sense to optimize this to only run once. For my open source projects I have it run twice.

For my understanding, this change will make it run twice but only on synchronize events which seem to be fired when the pull request branch is updated but not the base branch from my understanding? 
https://github.community/t/what-is-a-pull-request-synchronize-event/14784/4

I think its all fine by me. We can tweak it later.

runningcode (approved):

jthurne:
> For my understanding, this change will make it run twice but only on synchronize events which seem to be fired when the pull request branch is updated but not the base branch from my understanding?

That's my interpretation as well. Which I think is fine since we don't automatically merge in the base branch.

If we're worried about bugs sneaking in when we merge to the base branch, we could configure the repo to require PRs to be up-to-date with the base branch before allowing them to be merged. This seems like "quiet" enough of a repo that such a requirement wouldn't really slow development down any.  But that can be done outside of this PR.

PR #91, opened 2021-05-20
Add support for configuring Gradle Enterprise settings via system properties

jthurne:
The Gradle Enterprise Maven extension allows users to configure most Gradle
Enterprise configuration settings via Java system properties. This is really
nice when it comes time to run build experiments (particularly during a Gradle
Enterprise trial) because it makes it easy to change settings when running
builds on CI servers (it often avoids the need to make code changes that need
to be pushed to the git repo).

For example, it is possible to disable to local cache but enable the remote
cache through system properties when doing Maven Trial Experiment Number 3. The
user can configure _just_ their CI configuration to make this change and they
don't have to actually edit their Gradle Enterprise configuration files.

Unfortunately, the Gradle Build Tool and Gradle Enterprise Gradle Plugin only
support configuring a small number of settings via System properties.

To afford the same convenience to Gradle users, this PR updates the Common
Custom User Data Gradle plugin to honor the same (or similar) system properties
honored by the Gradle Enterprise Maven extension.

The same property names are used to make it easy for a user familiar with the
Maven extension to also configure a Gradle build (and vice versa).

### DIFF
--- a/common-custom-user-data-gradle-plugin/README.md
+++ b/common-custom-user-data-gradle-plugin/README.md
@@ -16,7 +16,7 @@ Out of the box, this plugin enhances all build scans published with additional t
 - For `git` repositories, information on the commit id, branch name, status, and whether the checkout is dirty or not.
 
 Out of the box, this plugin also allows users to override various Gradle Enterprise, build scan, and build caching settings via
-system properties.
+system properties (see [Overriding Gradle Enterprise configuration values via system properties](#overriding_gradle_enterprise_configuration_values_via_system_properties)).
 
 ### Applying the published plugin
 
@@ -81,6 +81,7 @@ The Gradle Enterprise plugin directly supports two other system properties:
 |`scan.captureTaskInputFiles`               | `true` or `false`   | Enables/disables capturing task input files in build scans.
 |`scan.uploadInBackground`                  | `true` or `false`   | Enables/disables uploading build scans in the background.
 
+See the [Gradle Enterprise Gradle Plugin User Manual](https://docs.gradle.com/enterprise/gradle-plugin/) for more details.
 
 ### Developing a customized version of the plugin
 
--- a/common-custom-user-data-gradle-plugin/README.md
+++ b/common-custom-user-data-gradle-plugin/README.md
@@ -15,6 +15,9 @@ Out of the box, this plugin enhances all build scans published with additional t
 - A tag representing builds run on CI, together with a set of tags, links and custom values specific to the CI server running the build.
 - For `git` repositories, information on the commit id, branch name, status, and whether the checkout is dirty or not.
 
+Out of the box, this plugin also allows users to override various Gradle Enterprise, build scan, and build caching settings via
+system properties.
+
 ### Applying the published plugin
 
 The Common Custom User Data Gradle Plugin is available in the [Gradle Plugin Portal](https://plugins.gradle.org/plugin/com.gradle.common-custom-user-data-gradle-plugin). This plugin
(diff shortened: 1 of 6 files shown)


jthurne:
I put this PR In draft mode because:

  1. I have not yet finished fully testing that all of this works. I've tested enough to know that the various config values show as having been properly set by inspecting build scans, but I haven't yet finished testing if they actually take affect.
  2. Need to add some details about all of these config settings to the README.

common-custom-user-data-gradle-plugin/src/main/java/com/gradle/CustomGradleEnterpriseConfig.java
```diff
@@ -50,9 +76,45 @@ static void configureBuildCache(BuildCacheConfiguration buildCache) {
...
+            buildCache.remote(HttpBuildCache.class).setPush(value);
+        });
+        withBooleanSystemProp(REMOTE_CACHE_ALLOW_UNTRUSTED_SERVER_PROP, value -> {
+            buildCache.remote(HttpBuildCache.class).setAllowUntrustedServer(value);
+        });
     }
 
-    private CustomGradleEnterpriseConfig() {
+    private void withSystemProp(String systemPropertyName, Consumer<String> action) {
+        Provider<String> prop = providers.systemProperty(systemPropertyName).forUseAtConfigurationTime();
```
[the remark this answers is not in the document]
jthurne: Good feedback!

1. This is done. I left some wrapper functions behind in CustomGradleEnterpriseConfig so that we don't have to pass the ProvidersFactory in every single call to withSystemProp and withBooleanSystemProp.

2. Thankfully, I have a Gradle 5.x project I can use to test this. I tested it, and it indeed does not work. Luckily, I think I can make it work with both Gradle 5.x and the configuration cache.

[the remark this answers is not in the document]
jthurne: I just pushed a commit that fixes this to work on earlier versions of Gradle as well.

common-custom-user-data-gradle-plugin/src/main/java/com/gradle/Utils.java
```diff
@@ -36,6 +42,28 @@ static boolean sysPropertyKeyStartingWith(String keyPrefix) {
         return false;
     }
 
+    static void withSystemProp(ProviderFactory providers, String systemPropertyName, Consumer<String> action) {
+        try {
```
runningcode, line 46: We usually do this by checking for the Gradle version instead of a `try`/`catch` block. I think it makes it a bit cleaner: https://github.com/gradle/gradle-enterprise-build-config-samples/blob/master/common-custom-user-data-gradle-plugin/src/main/java/com/gradle/CommonCustomUserDataGradlePlugin.java#L15

Also thanks for the quick fix!

common-custom-user-data-gradle-plugin/src/main/java/com/gradle/CustomGradleEnterpriseConfig.java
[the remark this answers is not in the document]
runningcode: Looks great! Thanks.

etiennestuder:
@jthurne I'll take a look today. one thing we need to change for sure is reading the sys props (and env vars?) in a way that is config cache compatible (meaning we need to use CC APIs to read/declare them).

common-custom-user-data-gradle-plugin/src/main/java/com/gradle/Utils.java
runningcode: Looks good.

common-custom-user-data-gradle-plugin/src/main/java/com/gradle/CommonCustomUserDataGradlePlugin.java
```diff
@@ -51,10 +68,16 @@ private void applyProjectPlugin(Project project) {
             CustomBuildScanEnhancements.configureBuildScan(buildScan, project.getGradle());
 
             // Build cache configuration cannot be accessed from a project plugin
+
+            project.afterEvaluate(___ -> {
```
runningcode, line 72: I know there is a slight performance penalty for using `afterEvaluate`. Why do we need it here? Could you add a comment to the code?

runningcode (approved):
Looks good to me. Also nice documentation.

common-custom-user-data-gradle-plugin/README.md
```diff
@@ -15,6 +15,9 @@ Out of the box, this plugin enhances all build scans published with additional t
 - A tag representing builds run on CI, together with a set of tags, links and custom values specific to the CI server running the build.
 - For `git` repositories, information on the commit id, branch name, status, and whether the checkout is dirty or not.
 
+Out of the box, this plugin also allows users to override various Gradle Enterprise, build scan, and build caching settings via
```
runningcode, line 18: You could also add a markdown link to the documentation for this here.

runningcode:
Pro tip for reviewers, you can add `?w=1` to the url to ignore the whitespace changes.

common-custom-user-data-gradle-plugin/src/main/java/com/gradle/CommonCustomUserDataGradlePlugin.java
jthurne: There is a comment for this, but it's within the block. I moved it out.

common-custom-user-data-gradle-plugin/README.md
jthurne: Good idea. Done!

common-custom-user-data-gradle-plugin/src/main/java/com/gradle/CommonCustomUserDataGradlePlugin.java
runningcode: AH ok, i didn't realize that was the reason for this block. But now im curious since the comment is the same, why  do we need to do this in both `settingsEvaluated` and `afterEvaluate`? Would it not be enough to do one or the other? Is this a Gradle 5 vs Gradle 6 thing?

runningcode (approved):

common-custom-user-data-gradle-plugin/src/main/java/com/gradle/CommonCustomUserDataGradlePlugin.java
jthurne: Yeah, it's a Gradle 5 vs Gradle 6+ thing.

For Gradle 6+, this is done in the context of settings.gradle(.kts).

For Gradle 5, this is done in the context of a build.gradle(.kts) script, at which point, Settings.settingsEvaluated doesn't really make sense.

Issue #92, opened 2021-05-26
Undeclared system property reads in common-custom-user-data-gradle-plugin


ALikhachev:
The Gradle plugin makes direct system property reads that may cause configuration cache issues, for example `idea.version` if executing build from IDEA with the plugin applied.

The plugin is used in [Kotlin build](https://github.com/JetBrains/kotlin/blob/b3bb033fe40a50246e5258a4a94313958a6636a5/settings.gradle#L44) and prevents configuration cache adoption alongside with Gradle Enterprise.

etiennestuder:
Agreed. We will get to it soon.

jthurne:
Fixed in the Common Custom User Data Gradle Plugin version 1.4, which we released today.

PR #93, opened 2021-05-27
Declare all environment variable and Gradle property access to support the configuration cache

jthurne:
We want the Common Custom User Data Gradle Plugin to be compatible with as many builds as possible. This includes Gradle builds which enable configuration caching.

Fixes Issue #92

### DIFF
--- a/common-custom-user-data-gradle-plugin/build.gradle
+++ b/common-custom-user-data-gradle-plugin/build.gradle
@@ -49,9 +49,10 @@ pluginBundle {
 
 signing {
     // Require publications to be signed on CI. Otherwise, publication will be signed only if keys are provided.
-    required System.getenv().containsKey("CI")
+    required providers.environmentVariable("CI").forUseAtConfigurationTime().isPresent()
 
-    useInMemoryPgpKeys(System.getenv("PGP_SIGNING_KEY"), System.getenv("PGP_SIGNING_KEY_PASSPHRASE"))
+    useInMemoryPgpKeys(providers.environmentVariable("PGP_SINGING_KEY").forUseAtConfigurationTime().orNull,
+        providers.environmentVariable("PGP_SIGNING_KEY_PASSPHRASE").forUseAtConfigurationTime().orNull)
 
     sign(configurations.archives)
 }
--- a/common-custom-user-data-gradle-plugin/build.gradle
+++ b/common-custom-user-data-gradle-plugin/build.gradle
@@ -39,7 +39,7 @@ When using this project as a template for your own plugin to publish internally,
 You may also remove `plugin-publish`, `signing` and `grgit` from the `plugins {}` block above.
  */
 
-version = System.getenv("RELEASE_OVERRIDE_VERSION") ?: version
+version = providers.environmentVariable("RELEASE_OVERRIDE_VERSION").forUseAtConfigurationTime().getOrElse(version)
 
 pluginBundle {
     website = "https://github.com/gradle/gradle-enterprise-build-config-samples/tree/master/common-custom-user-data-gradle-plugin"
--- a/common-custom-user-data-gradle-plugin/build.gradle
+++ b/common-custom-user-data-gradle-plugin/build.gradle
@@ -51,7 +51,8 @@ signing {
     // Require publications to be signed on CI. Otherwise, publication will be signed only if keys are provided.
     required providers.environmentVariable("CI").forUseAtConfigurationTime().isPresent()
 
-    useInMemoryPgpKeys(providers.environmentVariable("PGP_SINGING_KEY").forUseAtConfigurationTime().orNull,
+    useInMemoryPgpKeys(
+        providers.environmentVariable("PGP_SINGING_KEY").forUseAtConfigurationTime().orNull,
         providers.environmentVariable("PGP_SIGNING_KEY_PASSPHRASE").forUseAtConfigurationTime().orNull)
 
     sign(configurations.archives)
--- a/common-custom-user-data-gradle-plugin/src/main/java/com/gradle/CustomBuildScanEnhancements.java
+++ b/common-custom-user-data-gradle-plugin/src/main/java/com/gradle/CustomBuildScanEnhancements.java
@@ -33,7 +33,7 @@ final class CustomBuildScanEnhancements {
         captureOs(buildScan, providers);
         captureIde(buildScan, gradle, providers);
         captureCiOrLocal(buildScan, providers);
-        captureCiMetadata(buildScan, gradle);
+        captureCiMetadata(buildScan, gradle, providers);
         captureGitMetadata(buildScan);
         captureTestParallelization(buildScan, gradle);
     }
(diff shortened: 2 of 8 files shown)


common-custom-user-data-gradle-plugin/src/main/java/com/gradle/CustomBuildScanEnhancements.java
```diff
@@ -27,67 +29,59 @@
...
-        sysProperty("os.name").ifPresent(buildScan::tag);
-    }
-
-    private static void captureIde(BuildScanExtension buildScan, Gradle gradle) {
+    private static void captureIde(BuildScanExtension buildScan, Gradle gradle, ProviderFactory providers) {
         // Wait for projects to load to ensure Gradle project properties are initialized
         gradle.projectsEvaluated(g -> {
-            Project project = g.getRootProject();
-            if (project.hasProperty("android.injected.invoked.from.ide")) {
+            if (projectProperty(gradle,"android.injected.invoked.from.ide", providers).isPresent()) {
```
runningcode, line 48:
```suggestion
            if (projectProperty(gradle, "android.injected.invoked.from.ide", providers).isPresent()) {
```

common-custom-user-data-gradle-plugin/src/main/java/com/gradle/Utils.java
```diff
@@ -28,7 +29,11 @@ static boolean isNotEmpty(String value) {
         return value != null && !value.isEmpty();
     }
 
-    static Optional<String> sysProperty(String name) {
+    static Optional<String> sysProperty(String name, ProviderFactory providers) {
```
runningcode, line 32: This `Utils` file would likely make a nice library that we could give to plugin authors wanting to support both Gradle 5,6 and 7 builds and be compatible with the configuration cache.

runningcode (approved):
Looks good. I think the changes weren't too bad since all the usages of these properties were centrialized in `Utils`. Thanks Jim!

common-custom-user-data-gradle-plugin/src/main/java/com/gradle/Utils.java
jthurne: In the direction of a library, what might be particularly nice is to create a wrapper class (decorator) around ProviderFactory that handles the particulars of Gradle cross-version support.

runningcode:
I was just thinking that this makes the code sharing between the maven and gradle build a bit more complicated since we can't copy pasta anymore :(

common-custom-user-data-gradle-plugin/src/main/java/com/gradle/CustomBuildScanEnhancements.java
```diff
@@ -17,6 +18,7 @@
 import static com.gradle.Utils.execAndCheckSuccess;
 import static com.gradle.Utils.execAndGetStdOut;
 import static com.gradle.Utils.isNotEmpty;
+import static com.gradle.Utils.projectProperty;
 import static com.gradle.Utils.readPropertiesFile;
```
etiennestuder, line 22: I believe how we read this file is probably not config cache correct.

common-custom-user-data-gradle-plugin/src/main/java/com/gradle/CustomBuildScanEnhancements.java
```diff
@@ -27,67 +29,59 @@
...
+    private static void captureOs(BuildScanExtension buildScan, ProviderFactory providers) {
+        sysProperty("os.name", providers).ifPresent(buildScan::tag);
     }
 
-    private static void captureOs(BuildScanExtension buildScan) {
-        sysProperty("os.name").ifPresent(buildScan::tag);
-    }
-
-    private static void captureIde(BuildScanExtension buildScan, Gradle gradle) {
+    private static void captureIde(BuildScanExtension buildScan, Gradle gradle, ProviderFactory providers) {
```
etiennestuder, line 45: I believe you can refactor these if/else into ifPresent

etiennestuder (changes requested):

common-custom-user-data-gradle-plugin/build.gradle
```diff
@@ -39,7 +39,7 @@ When using this project as a template for your own plugin to publish internally,
 You may also remove `plugin-publish`, `signing` and `grgit` from the `plugins {}` block above.
  */
 
-version = System.getenv("RELEASE_OVERRIDE_VERSION") ?: version
+version = providers.environmentVariable("RELEASE_OVERRIDE_VERSION").forUseAtConfigurationTime().getOrElse(version)
 
 pluginBundle {
     website = "https://github.com/gradle/gradle-enterprise-build-config-samples/tree/master/common-custom-user-data-gradle-plugin"
```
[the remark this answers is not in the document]
jthurne: Good catch. Fixed.

common-custom-user-data-gradle-plugin/src/main/java/com/gradle/SystemPropertyOverrides.java
```diff
@@ -27,24 +27,24 @@
...
         if (buildCache.getRemote() != null) {
             buildCache.remote(HttpBuildCache.class, remote -> {
-                withSysProperty(REMOTE_CACHE_URL, remote::setUrl, providers);
-                withBooleanSysProperty(REMOTE_CACHE_ALLOW_UNTRUSTED_SERVER, remote::setAllowUntrustedServer, providers);
-                withBooleanSysProperty(REMOTE_CACHE_ENABLED, remote::setEnabled, providers);
-                withBooleanSysProperty(REMOTE_CACHE_PUSH_ENABLED, remote::setPush, providers);
+                sysProperty(REMOTE_CACHE_URL, providers).ifPresent(remote::setUrl);
+                booleanSysProperty(REMOTE_CACHE_ALLOW_UNTRUSTED_SERVER, providers).ifPresent(remote::setAllowUntrustedServer);
+                booleanSysProperty(REMOTE_CACHE_ENABLED, providers).ifPresent(remote::setEnabled);
+                booleanSysProperty(REMOTE_CACHE_PUSH_ENABLED, providers).ifPresent(remote::setPush);
```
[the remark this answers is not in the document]
jthurne: I merged master into this branch and resolved the conflict. Thanks for the heads-up!

common-custom-user-data-gradle-plugin/src/main/java/com/gradle/CustomBuildScanEnhancements.java
jthurne: It is possible, but getting the same semantics is tricky. I'm not sure the `ifPresent` version is better/clearer than the `if-else if` version. Here's the best that I could come up with:

```java
final AtomicReference<String> ideTag = new AtomicReference<>("");
projectProperty("android.injected.invoked.from.ide", providers, gradle).ifPresent(__ -> ideTag.set("Android Studio"));
projectProperty("android.injected.studio.version", providers, gradle).ifPresent(v -> buildScan.value("Android Studio version", v));

sysProperty("idea.version", providers).ifPresent(__ -> ideTag.set("IntelliJ IDEA"));
sysProperty("eclipse.buildId", providers).ifPresent(__ -> ideTag.set("Eclipse"));

if(sysPropertyKeyStartingWith("idea.version")) {
    ideTag.set("IntelliJ IDEA");
}
if(!isCi(providers)) {
    ideTag.set("Cmd Line");
}

if (!ideTag.get().isEmpty()) {
    buildScan.tag(ideTag.get());
}
```

If we think it's better, then I'll gladly change the code to the above. Other suggestions are also welcome.

I thought about using `Optional.ifPresentOrElse`, but that is even more complex, and doesn't work for every case since some of the else if clauses contain regular boolean expressions (like `!isCi(providers`).

common-custom-user-data-gradle-plugin/src/main/java/com/gradle/Utils.java
```diff
@@ -92,6 +97,14 @@ static Properties readPropertiesFile(String name) {
         }
     }
 
+    static String readFile(String name, ProviderFactory providers, Gradle gradle) {
+        Provider<File> file = providers.provider(() -> new File(name));
```
runningcode: 🙈

etiennestuder: What do you mean by this comment, @runningcode ?

common-custom-user-data-gradle-plugin/src/main/java/com/gradle/CustomBuildScanEnhancements.java
etiennestuder: Let's leave the current if/else structure.

etiennestuder:
@jthurne as discussed, once the ide versions are captured as custom values, this PR is good to be merged to master.

runningcode:
🥳

PR #94, opened 2021-06-01
Check if on CI before capturing IDE details in the CCUDME

jthurne:
We are already doing this check in the Gradle plugin.

### DIFF
--- a/common-custom-user-data-maven-extension/src/main/java/com/gradle/CustomBuildScanEnhancements.java
+++ b/common-custom-user-data-maven-extension/src/main/java/com/gradle/CustomBuildScanEnhancements.java
@@ -40,14 +40,16 @@ final class CustomBuildScanEnhancements {
     }
 
     private static void captureIde(BuildScanApi buildScan) {
-        if (sysProperty("idea.version").isPresent()) {
-            buildScan.tag("IntelliJ IDEA");
-            buildScan.value("IntelliJ IDEA version", sysProperty("idea.version").get());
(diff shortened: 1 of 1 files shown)


common-custom-user-data-maven-extension/src/main/java/com/gradle/CustomBuildScanEnhancements.java
```diff
@@ -40,14 +40,16 @@ private static void captureOs(BuildScanApi buildScan) {
...
-        } else if (!isCi()) {
-            buildScan.tag("Cmd Line");
+        if (!isCi()) {
+            if (sysProperty("idea.version").isPresent()) {
+                buildScan.tag("IntelliJ IDEA");
+                buildScan.value("IntelliJ IDEA version", sysProperty("idea.version").get());
+            } else if (sysProperty("eclipse.buildId").isPresent()) {
+                buildScan.tag("Eclipse");
+                buildScan.value("Eclipse version", sysProperty("eclipse.buildId").get());
+            } else if (!isCi()) {
```
etiennestuder, line 50: @jthurne this check is now redundant. (see the gradle plugin for comparison)

etiennestuder: Once fixed, you can merge to master.

jthurne: > @jthurne this check is now redundant.

It is indeed! Fixed.

PR #95, opened 2021-06-01
Fix reading project properties with Gradle 6.5 or higher

jthurne:
`ProviderFactory.gradleProperty` currently does not read local project properties
(see https://github.com/gradle/gradle/issues/13302). This commit attempts to
use a workaround: Read the property within a callable that is used to create a
Provider.

Related to #93

### DIFF
--- a/common-custom-user-data-gradle-plugin/src/main/java/com/gradle/Utils.java
+++ b/common-custom-user-data-gradle-plugin/src/main/java/com/gradle/Utils.java
@@ -56,7 +56,7 @@ final class Utils {
 
     static Optional<String> projectProperty(String name, ProviderFactory providers, Gradle gradle) {
         if (isGradle65OrNewer()) {
-            Provider<String> property = providers.gradleProperty(name).forUseAtConfigurationTime();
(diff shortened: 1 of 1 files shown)


runningcode (approved):
Surprising behavior. Thanks for fixing.

etiennestuder:
@jthurne Why do we need support for non-root project properties?

jthurne:
> Why do we need support for non-root project properties?

It looks like the init script that IntelliJ uses to create those properties creates them on the root project. But ProviderFactory.gradleProperty apparently doesn't have access to them (see gradle/gradle#13302), regardless if it is the root project or not.

I think ProviderFactory.gradleProperty is only able to access properties that are set at the commandline via `-P`.

etiennestuder:
> ProviderFactory.gradleProperty apparently doesn't have access to them (see gradle/gradle#13302), regardless if it is the root project or not.

Ok. I now get it. From the Gradle BT issue, I had the impression it all works with any root project props.

jthurne:
I just realized this fix may not properly invalidate the configuration cache if the property value changes. I'll give this one some more thought.

PR #96, opened 2021-06-02
Reintroduce support for older IntelliJ IDEA versions

jthurne:
For Gradle, the new configuration-cache-compatible solution uses the fact that
there should only ever be one system property that starts with `idea.version`.
If one exists, then the Plugin lets Gradle know that it should include the
property when calculating the configuration cache key. If the user upgrades
IntelliJ, then the previous system property will be removed and the
configuration cache will be invalidated. If a new property takes its place,
then that one will be declared.

For both Maven and Gradle, if an `idea.versionYYYY.NN` System property is
found, then the System property's key is parsed to populate the "IntelliJ IDEA
version" custom value.

Related to #92

### DIFF
--- a/common-custom-user-data-gradle-plugin/src/main/java/com/gradle/Utils.java
+++ b/common-custom-user-data-gradle-plugin/src/main/java/com/gradle/Utils.java
@@ -48,10 +48,10 @@ final class Utils {
 
     private static Optional<String> firstKeyStartingWith(String keyPrefix, Properties properties) {
         return properties.keySet().stream()
-                .filter(s -> s instanceof String)
-                .map(s -> (String) s)
-                .filter(s -> s.startsWith(keyPrefix))
-                .findFirst();
+            .filter(s -> s instanceof String)
+            .map(s -> (String) s)
+            .filter(s -> s.startsWith(keyPrefix))
+            .findFirst();
     }
 
     static Optional<Boolean> booleanSysProperty(String name, ProviderFactory providers) {
@@ -82,6 +82,10 @@ final class Utils {
         return str.endsWith(suffix) ? str : str + suffix;
     }
 
+    static String stripPrefix(String prefix, String string) {
+        return string.startsWith(prefix) ? string.substring(prefix.length()) : string;
+    }
+
     static String urlEncode(String str) {
         try {
             return URLEncoder.encode(str, StandardCharsets.UTF_8.name());
@@ -101,7 +105,7 @@ final class Utils {
     }
 
     static InputStream readFile(String name, ProviderFactory providers, Gradle gradle) throws FileNotFoundException {
-        if(isGradle65OrNewer()) {
+        if (isGradle65OrNewer()) {
             RegularFile file = gradle.getRootProject().getLayout().getProjectDirectory().file(name);
             Provider<byte[]> fileContent = providers.fileContents(file).getAsBytes().forUseAtConfigurationTime();
 
@@ -170,4 +174,5 @@ final class Utils {
 
     private Utils() {
     }
+
 }
--- a/common-custom-user-data-gradle-plugin/src/main/java/com/gradle/Utils.java
+++ b/common-custom-user-data-gradle-plugin/src/main/java/com/gradle/Utils.java
@@ -41,7 +41,7 @@ final class Utils {
         Optional<String> key = firstKeyStartingWith(keyPrefix, System.getProperties());
 
         if (isGradle65OrNewer()) {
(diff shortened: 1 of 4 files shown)


common-custom-user-data-gradle-plugin/src/main/java/com/gradle/Utils.java
```diff
@@ -38,6 +37,23 @@ static boolean isNotEmpty(String value) {
         return Optional.ofNullable(System.getProperty(name));
     }
 
+    static Optional<String> firstSysPropertyKeyStartingWith(String keyPrefix, ProviderFactory providers) {
+        Optional<String> key = firstKeyStartingWith(keyPrefix, System.getProperties());
```
runningcode, line 41: Just to confirm: checking `System.getProperties()` is okay for config cache?
I think a javadoc on this method might be helpful to customers modifying this plugin for their own needs.

runningcode (approved):
just a comment about documentation. otherwise looks good for me. nice find!

common-custom-user-data-gradle-plugin/src/main/java/com/gradle/Utils.java
jthurne: > Just to confirm: checking System.getProperties() is okay for config cache?

Surprisingly, yes. I don't know what kind of magic the Configuration Cache is using to detect when there is undeclared access to system properties, but it does appear to be able to determine that there isn't any undeclared access going on.  Maybe it intercepts calls to System.getProperties().get() or something.

jthurne: > I think a javadoc on this method might be helpful to customers modifying this plugin for their own needs.

Good idea!

common-custom-user-data-maven-extension/src/main/java/com/gradle/CustomBuildScanEnhancements.java
```diff
@@ -47,12 +48,19 @@ private static void captureIde(BuildScanApi buildScan) {
             } else if (sysProperty("eclipse.buildId").isPresent()) {
                 buildScan.tag("Eclipse");
                 buildScan.value("Eclipse version", sysProperty("eclipse.buildId").get());
+            } else if (firstSysPropertyKeyStartingWith("idea.version").isPresent()) {
+                buildScan.tag("IntelliJ IDEA");
+                buildScan.value("IntelliJ IDEA version", stripPrefix("idea.version", firstSysPropertyKeyStartingWith("idea.version").get()));
```
etiennestuder, line 53: I don't understand how this logic works for newer versions of IDEA. there is no code here that reads the actual value of the sys prop idea.version. Or am I missing something?

etiennestuder (changes requested):

common-custom-user-data-maven-extension/src/main/java/com/gradle/CustomBuildScanEnhancements.java
etiennestuder: I now realize that there are two IDEA blocks. I think they can be merged into one as it was before.

common-custom-user-data-maven-extension/src/main/java/com/gradle/CustomBuildScanEnhancements.java
```diff
@@ -47,12 +48,19 @@ private static void captureIde(BuildScanApi buildScan) {
...
             } else {
                 buildScan.tag("Cmd Line");
             }
         }
     }
 
+    private static String stripPrefix(String prefix, String string) {
+        return string.startsWith(prefix) ? string.substring(prefix.length()) : string;
+    }
```
etiennestuder, line 62: Move to Utils class, for both gradle and maven.

etiennestuder (changes requested):

common-custom-user-data-maven-extension/src/main/java/com/gradle/CustomBuildScanEnhancements.java
etiennestuder: I think the problem here is that in case idea.version sys prop is present, both IDEA blocks will be executed.

etiennestuder: No, it is not, since there is an if/else statement.

etiennestuder (approved):
I did some small code refactors. LGTM to merge now.

etiennestuder (approved):

PR #97, opened 2021-06-07
Adjust reading project properties, capture full Git commit ID, and add another system property override

jthurne:
See commit messages for details.

Related to #92 .


common-custom-user-data-gradle-plugin/src/main/java/com/gradle/SystemPropertyOverrides.java
```diff
@@ -46,6 +51,9 @@ static void configureBuildCache(BuildCacheConfiguration buildCache, ProviderFact
         if (buildCache.getRemote() != null) {
             buildCache.remote(HttpBuildCache.class, remote -> {
                 sysProperty(REMOTE_CACHE_URL, providers).ifPresent(remote::setUrl);
+                if (remote.getUrl() != null) {
```
jthurne, line 54: I'm not sure if this check is strictly needed because it seems unlikely that a user would have a HttpBuildCache configured with a null URL.

etiennestuder (approved):

PR #98, opened 2021-06-07
Improve consistency between Gradle and Maven `CustomBuildScanEnhancements`

bigdaz:
For the Gradle and Maven custom-user-data plugin/extension, the build scan enhancement code is logically identical. However there are small differences between the implementation for Gradle and Maven, primarily to support the Gradle configuration-cache and late binding of project properties.

Over time, these implementations have diverged somewhat, making it more difficult to keep them in sync using a simple file comparison. Maintaining these divergent implementation is a bit tricky. We could find a way to extract all of the common code between these modules, but this would result in additional complexity for users who are using these projects as templates for their own extension/plugin.

Instead, this PR simply restructures the code a little to minimize the divergence between the 2 implementations. This permits a simple file copy in IDEA to be used to compare the 2 implementations.

Before this change, the comparison looks a bit like this:
image
image

But after this cleanup, the comparison is easier to grok:
image
image


etiennestuder:
I like the pattern. Now that the changes to the plugin & extension have stabilized, it would be a good time to add this pattern but to start over given the divergence of this branch and master. @jthurne do you think you can take than on?

jthurne:
Updated and resubmitted as #103

PR #99, opened 2021-06-08
Add GradleEnterpriseApi

guylabs:
This originates from https://github.com/gradle/dotcom/issues/9643

Once merged these tests must be unignored: https://github.com/gradle/dotcom/blob/aeed93e05526490ba83893825d116aa8ca56999f/common-custom-user-data-maven-extension-test-func/src/test/groovy/com/gradle/maven/custom/CommonCustomUserDataMavenExtensionFuncTest.groovy#L239


common-custom-user-data-maven-extension/README.md
```diff
@@ -18,6 +18,15 @@ The additional tags, links and custom values captured by this extension include:
...
 captured and under which conditions.
 
+#### Version compatibility
+
+This table details the version compatibility of the Common Custom User Data Maven extension with the Gradle Enterprise Maven extension.
+
+| Common Custom User Data Maven extension versions | Gradle Enterprise Maven extension versions |
+| ------------------------------------------------ | ------------------------------------------ |
+| `1.5.1`                                          | `1.10.1+`                                  |
+| `1.0` - `1.5.0`                                  | `1.0.0` - `1.10`                           |
```
etiennestuder, line 28: is it really the case that i cannot use CCUD extension 1.5.0 with GE extension 1.10.1?

etiennestuder (commented):
Left one comment. Everything else LGTM.

I can marge this later this week once we have released a version of the CCUD Maven extension with all the other changes that have accumulated.

I will then also update the version table that was added in this PR.

common-custom-user-data-maven-extension/README.md
facewindu, line 28: This is not true.
The build cache API was introduced with version 1.6.5 according to the javadoc.
BuildScanApi was introduced with 1.0 (and it's 1.0, not 1.0.0), but some things on it came later (not sure if we want / need to go into details here, maybe just mention to look at https://docs.gradle.com/enterprise/maven-extension/api)

facewindu (changes requested):

common-custom-user-data-maven-extension/README.md
guylabs: True. Fixed.

guylabs: Updated this accordingly.

common-custom-user-data-maven-extension/README.md
```diff
@@ -18,6 +18,15 @@ The additional tags, links and custom values captured by this extension include:
 See [CustomBuildScanEnhancements.java](./src/main/java/com/gradle/CustomBuildScanEnhancements.java) for details on what data is
 captured and under which conditions.
 
+#### Version compatibility
+
+This table details the version compatibility of the Common Custom User Data Maven extension with the Gradle Enterprise Maven extension.
+
+| Common Custom User Data Maven extension versions | Gradle Enterprise Maven extension versions |
+| ------------------------------------------------ | ------------------------------------------ |
+| `1.5.1`                                          | `1.10.1+`                                  |
```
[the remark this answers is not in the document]
guylabs: Fixed.

facewindu (approved):

common-custom-user-data-maven-extension/README.md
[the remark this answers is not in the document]
etiennestuder: how so? it should be 1.6. or maybe I do not see the commit that fixed it.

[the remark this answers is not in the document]
guylabs: https://github.com/gradle/gradle-enterprise-build-config-samples/pull/99/commits/c86f7bc0c57445282f0f07e0453669fda8b4a517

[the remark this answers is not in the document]
etiennestuder: What I meant is that we will not release 1.5.1 but a new minor version:
- 1.6 with the current changes on master
- 1.7 with the change of this PR
I will update the table accordingly once we get there.

common-custom-user-data-maven-extension/src/main/java/com/gradle/CommonCustomUserDataMavenExtension.java
```diff
@@ -31,6 +32,13 @@ public CommonCustomUserDataMavenExtension(PlexusContainer container, Logger logg
...
         logger.debug("Executing extension: " + getClass().getSimpleName());
 
+        GradleEnterpriseApi gradleEnterprise = ApiAccessor.lookupGradleEnterpriseApi(container, getClass());
+        if (gradleEnterprise != null) {
+            logger.debug("Configuring Gradle Enterprise");
+            CustomGradleEnterpriseConfig.configureGradleEnterprise(gradleEnterprise);
+            logger.debug("Finished configuring Gradle Enterprise");
+        }
+
         BuildScanApi buildScan = ApiAccessor.lookupBuildScanApi(container, getClass());
```
marcphilipp, line 42: I think we should remove these additional lookups. @etiennestuder Do you think we should keep the separate methods in `CustomGradleEnterpriseConfig`?

etiennestuder: Hm, I think I would keep it as it is. But I can take another look when I merge it.

What is your rationale for changing it?

marcphilipp: It would emphasize that there's a single entrypoint but I'm okay with leaving it as-is.

facewindu: If you remove it, then it breaks existing customer scripts

marcphilipp: I wasn't suggesting to remove it from the Groovy script support, just from the Java "templates".

common-custom-user-data-maven-extension/README.md
etiennestuder: @guylabs it's not clear to me why we need such a commpatibility table? why can I not use CCUD Maven extension 1.6 with GE Maven extension 1.9.2? (I understand that we would not get back anything for GradleEnterpriseApi), but that does not mean they are not compatible.

etiennestuder: //cc @facewindu

facewindu: Because of the import of the various API interfaces. Right on `CommonCustomUserDataMavenExtension`, it references them so in your example, it would not work at runtime, because not finding the import for the GradleEnterpriseApi interface.
For it to work, you'd need to return an `Object` from `ApiAccessor.lookupGradleEnterpriseApi`, and cast it back to its proper type, when not null, and when you really need it (in eg `GroovyScriptUserData.addToApis`)

etiennestuder: Right. I fogot that the GE Maven extension can be a compileOnly dependency, as in our build by default.

common-custom-user-data-maven-extension/README.md
```diff
@@ -18,6 +18,16 @@ The additional tags, links and custom values captured by this extension include:
...
+| `1.3` - `1.5.0`                                  | `1.6.5+`                                   |
+| `1.0` - `1.3`                                    | `1.0+`                                     |
```
etiennestuder, line 29: @facewindu 1.3 appears in both rows? Is that correct? it seems to me that it that the last line should be `1.0 - 1.2`.

| `1.7+`                                              | `1.10.1+`                                 |
| `1.3` - `1.6`                                    | `1.6.5+`                                  |
| `1.0` - `1.2`                                    | `1.0+`                                     |

common-custom-user-data-maven-extension/src/main/java/com/gradle/CommonCustomUserDataMavenExtension.java
etiennestuder: I now see what you mean, @marcphilipp . I'll adjust the code accordingly. Thanks.

common-custom-user-data-maven-extension/README.md
etiennestuder: I fixed the table. Resolving now.

etiennestuder: I fixed the table. Resolving now.

common-custom-user-data-maven-extension/README.md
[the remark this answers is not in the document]
etiennestuder: I fixed the table. Resolving now.

common-custom-user-data-maven-extension/src/main/java/com/gradle/CommonCustomUserDataMavenExtension.java
etiennestuder: Fixed in this [commit](https://github.com/gradle/gradle-enterprise-build-config-samples/commit/58f49dbaceb10a3500c24c77b6673b6dd41530f9).

PR #103, opened 2021-06-14
Convert CustomBuildScanEnhancements to a normal class to improve consistency

jthurne:
- Pass context as constructor parameters, to avoid having to pass different method parameters.
- Extracted non-static `sysProperty` `projectProperty` and `envVariable` methods to avoid
  differences in method parameters for most calls.
- Reordered methods for consistency

### DIFF
--- a/common-custom-user-data-gradle-plugin/src/main/java/com/gradle/CustomBuildScanEnhancements.java
+++ b/common-custom-user-data-gradle-plugin/src/main/java/com/gradle/CustomBuildScanEnhancements.java
@@ -83,15 +83,15 @@ final class CustomBuildScanEnhancements {
     private void captureCiMetadata() {
         if (isJenkins() || isHudson()) {
             envVariable("BUILD_URL").ifPresent(url ->
-                buildScan.link(isJenkins() ? "Jenkins build" : "Hudson build", url));
+                    buildScan.link(isJenkins() ? "Jenkins build" : "Hudson build", url));
             envVariable("BUILD_NUMBER").ifPresent(value ->
-                buildScan.value("CI build number", value));
+                    buildScan.value("CI build number", value));
             envVariable("NODE_NAME").ifPresent(value ->
-                addCustomValueAndSearchLink("CI node", value));
+                    addCustomValueAndSearchLink("CI node", value));
             envVariable("JOB_NAME").ifPresent(value ->
-                addCustomValueAndSearchLink("CI job", value));
+                    addCustomValueAndSearchLink("CI job", value));
             envVariable("STAGE_NAME").ifPresent(value ->
-                addCustomValueAndSearchLink("CI stage", value));
+                    addCustomValueAndSearchLink("CI stage", value));
         }
 
         if (isTeamCity()) {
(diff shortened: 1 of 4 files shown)


jthurne:
@etiennestuder do we want to apply the same sort of refactoring to CustomGradleEnterpriseConfig and SystemPropertyOverrides? While it isn't necessary for consistency between the Gradle plugin and Maven extension, it would make the code within a single project more consistent.

common-custom-user-data-gradle-plugin/src/main/java/com/gradle/CustomBuildScanEnhancements.java
```diff
@@ -13,44 +13,49 @@
...
+    void apply() {
+        captureOs();
+        captureIde();
+        captureCiOrLocal();
+        captureCiMetadata();
+        captureGitMetadata();
+        captureTestParallelization();
     }
 
-    private static void captureOs(BuildScanExtension buildScan, ProviderFactory providers) {
```
runningcode, line 41: these were previously all static methods because of an issue being compatible with the configuration cache. I think this should just work now that they are members on the class but we should check before merging.

etiennestuder (approved):
Looking great.

etiennestuder:
> @etiennestuder do we want to apply the same sort of refactoring to CustomGradleEnterpriseConfig and SystemPropertyOverrides? While it isn't necessary for consistency between the Gradle plugin and Maven extension, it would make the code within a single project more consistent.

@jthurne I would merge this PR once you have ensured config caching is still supported (as requested by Nelson). And then create a separate PR to adjust the other Config classes. This makes it a bit easier to grasp the changes.

common-custom-user-data-gradle-plugin/src/main/java/com/gradle/CustomBuildScanEnhancements.java
jthurne: Good suggestion. It looks like these changes do cause a problem with the configuration cache. I'm exploring options.

jthurne: Configuration cache fix has been added to the PR.

common-custom-user-data-gradle-plugin/src/main/java/com/gradle/CustomBuildScanEnhancements.java
```diff
@@ -71,226 +76,219 @@ private static void captureIde(BuildScanExtension buildScan, ProviderFactory pro
...
+    private void captureGitMetadata() {
+        buildScan.background(captureGitMetadataAction);
+    }
 
-            String gitRepo = execAndGetStdOut("git", "config", "--get", "remote.origin.url");
-            String gitCommitId = execAndGetStdOut("git", "rev-parse", "--verify", "HEAD");
-            String gitCommitShortId = execAndGetStdOut("git", "rev-parse", "--short=8", "--verify", "HEAD");
-            String gitBranchName = execAndGetStdOut("git", "rev-parse", "--abbrev-ref", "HEAD");
-            String gitStatus = execAndGetStdOut("git", "status", "--porcelain");
+    private static final Action<BuildScanExtension> captureGitMetadataAction =
```
runningcode, line 233: ah i see, writing it in this way ensure that it doesn't capture the outer class and ensure it is compatible with the configuration cache. nice find. i wonder if we should document this in the code to avoid future incompatible changes or write a test for it.

runningcode (approved):
looks good to me!

common-custom-user-data-gradle-plugin/src/main/java/com/gradle/CustomBuildScanEnhancements.java
etiennestuder: Yes, that was the plan.

etiennestuder (commented):
Nice outcome.

PR #104, opened 2021-06-15
Use buildId instead of buildNumber for TeamCity

DonatasNasdaq:
buildNumber in TeamCity may be any String with whitespace characters. Thus it is better to use buildId instead.


jthurne:
Superseeded by #108

PR #105, opened 2021-06-16
Alternative fix for the configuration cache issue found in PR #103

jthurne:
Fix configuration cache support by making captureGitMetadata static

captureGitMetadata passes a function to BuildScanExtension.background. In order
to support configuration caching, that function needs to be serializable (the
serialized version of the function is stored in the configuration cache).

If captureGitMetadata is _not_ static, then the resulting function is actually
an anonymous inner class of CustomBuildScanEnhancements, which means the
serialized version also contains a serialized version of the
CustomBuildScanEnhancmeents instance (thus, CustomBuildScanEnhancmenets also
needs to be serializable). However, CustomBuildScanEnhancements has a reference
to a Gradle object, which is _not_ serializable. Thus, this makes
captureGitMetadata incompatible with configuration caching.

By making captureGitMetadata static, it is disassociated with any particular
instance of CustomBuildScanEnhancements, and therefore the lambda can be
serialized (and is therefore configuration cache compatible).

For consistency, the same modification is made to the CCUD Maven extension's
version of the class, even though Maven is not affected by this.


jthurne:
Incorporated this into #103

PR #106, opened 2021-06-16
Generate a "fat" jar for the CCUD Maven extension

jthurne:
In addition to adding the extension to the extensions.xml file, the CCUD Maven
extension can be used by putting it in the MAVEN_HOME/lib/ext directory, or by
referencing the jar on the command line by setting the `maven.ext.class.path`
system property when invoking Maven.

Unfortunately, if the extension is used via one of these two additonal methods,
then Maven is unable to resolve any dependencies of the CCUD Maven extension.
That means the extension could fail with a ClassDefNotFound error.

Users could add all of the extensions dependencies to their MAVEN_HOME/lib
folder, but we can make it easier on users if we publish a jar that contains
all of the extension's dependencies (which is currently only Groovy).

### DIFF
--- a/common-custom-user-data-maven-extension/pom.xml
+++ b/common-custom-user-data-maven-extension/pom.xml
@@ -114,6 +114,16 @@
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-shade-plugin</artifactId>
                 <version>3.2.4</version>
+                <configuration>
+                    <filters>
+                        <filter>
+                            <artifact>org.codehaus.groovy:groovy</artifact>
+                            <excludes>
+                                <exclude>META-INF/MANIFEST.MF</exclude>
+                            </excludes>
+                        </filter>
+                    </filters>
+                </configuration>
                 <executions>
                     <execution>
                         <phase>package</phase>
--- a/common-custom-user-data-maven-extension/pom.xml
+++ b/common-custom-user-data-maven-extension/pom.xml
@@ -110,6 +110,19 @@
                     </execution>
                 </executions>
             </plugin>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-shade-plugin</artifactId>
+                <version>3.2.4</version>
+                <executions>
+                    <execution>
+                        <phase>package</phase>
+                        <goals>
+                            <goal>shade</goal>
+                        </goals>
+                    </execution>
+                </executions>
+            </plugin>
         </plugins>
     </build>
 
--- a/common-custom-user-data-maven-extension/.gitignore
+++ b/common-custom-user-data-maven-extension/.gitignore
@@ -16,6 +16,7 @@ pom.xml.releaseBackup
 pom.xml.versionsBackup
 pom.xml.next
 release.properties
+dependency-reduced-pom.xml
 
 # OS X
 # ----
--- a/common-custom-user-data-maven-extension/CHANGELOG.md
+++ b/common-custom-user-data-maven-extension/CHANGELOG.md
@@ -2,6 +2,7 @@
 All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
+- Publish a "fat" jar that contains all of the extension's dependencies (to make it easier to use the extension in certian scenarios).
 
 ## [1.7.1] - 2021-06-22
 - Encode (escape) the TeamCity build number when creating a link to the TeamCity build (fixes Issue #101).


runningcode:
This seems like a good idea but this makes the extension a bit more heavyweight. I am unsure of the trade-off here of which one is best.

jthurne:
> This seems like a good idea but this makes the extension a bit more heavyweight.

That is true, but I think the usability improvement (not having to remember to copy the dependencies into MAVEN_HOME/lib) is probably worth it. It also reduces what we need to document/explain to users. And it makes it easier to maintain our internal functional tests for the extension.

etiennestuder:
@marcphilipp WDYT?

common-custom-user-data-maven-extension/pom.xml
```diff
@@ -110,6 +110,19 @@
...
+                <version>3.2.4</version>
+                <executions>
+                    <execution>
+                        <phase>package</phase>
+                        <goals>
+                            <goal>shade</goal>
+                        </goals>
+                    </execution>
+                </executions>
+            </plugin>
```
marcphilipp, line 125: I think this should not replace but attach the shaded artifact as an alternative artifact (cf. https://maven.apache.org/plugins/maven-shade-plugin/examples/attached-artifact.html).

marcphilipp (changes requested):
I think this is a good idea for the two use cases mentioned above. We don't have to change the `.mvn/extensions.xml` use case, though, if we publish a regular and a shaded version.

common-custom-user-data-maven-extension/pom.xml
runningcode: Oh, that is neat. That way we can benefit from both version. 👍

etiennestuder: What is the motivation for publishing two variants?
How will this look in practice for our users and our documentation?

marcphilipp: If we don't add it as an alternative artifact, Maven will use the regular POM with the shaded artifact and download the fat jar and all regular transitive dependencies. In the documentation we could link directly to the fat jar for the aforementioned use cases.

marcphilipp (approved):
I had a brief chat about this with @etienne. We agreed that publishing a single jar would be preferable as it is less confusing. My concern was that Maven would publish the project with the original POM which would lead to additionally downloading Groovy and its dependencies. I tried this locally now and at least for `mvn install`, the Shade plugin configures Maven to use the `dependency-reduced-pom.xml` that is generated by the `shade` goal:
```
[INFO] --- maven-install-plugin:2.4:install (default-install) @ common-custom-user-data-maven-extension ---
[INFO] Installing /Users/marc/Repositories/gradle/gradle-enterprise-build-config-samples/common-custom-user-data-maven-extension/target/common-custom-user-data-maven-extension-1.7.1-SNAPSHOT.jar to /Users/marc/.m2/repository/com/gradle/common-custom-user-data-maven-extension/1.7.1-SNAPSHOT/common-custom-user-data-maven-extension-1.7.1-SNAPSHOT.jar
[INFO] Installing /Users/marc/Repositories/gradle/gradle-enterprise-build-config-samples/common-custom-user-data-maven-extension/dependency-reduced-pom.xml to /Users/marc/.m2/repository/com/gradle/common-custom-user-data-maven-extension/1.7.1-SNAPSHOT/common-custom-user-data-maven-extension-1.7.1-SNAPSHOT.pom
```

While doing that I noticed a warning that we should solve by excluding `META-INF/MANIFEST.MF` from the Groovy JAR in the Shade plugin's configuration:
```
[INFO] --- maven-shade-plugin:3.2.4:shade (default) @ common-custom-user-data-maven-extension ---
[INFO] Including org.codehaus.groovy:groovy:jar:3.0.7 in the shaded jar.
[WARNING] common-custom-user-data-maven-extension-1.7.1-SNAPSHOT.jar, groovy-3.0.7.jar define 1 overlapping resource:
[WARNING]   - META-INF/MANIFEST.MF
[WARNING] maven-shade-plugin has detected that some class files are
[WARNING] present in two or more JARs. When this happens, only one
[WARNING] single version of the class is copied to the uber jar.
[WARNING] Usually this is not harmful and you can skip these warnings,
[WARNING] otherwise try to manually exclude artifacts based on
[WARNING] mvn dependency:tree -Ddetail=true and the above output.
[WARNING] See http://maven.apache.org/plugins/maven-shade-plugin/
```

jthurne:
> While doing that I noticed a warning that we should solve by excluding META-INF/MANIFEST.MF from the Groovy JAR in the Shade plugin's configuration:

This is done.

jthurne:
We will include this in the next release of the extension.

PR #108, opened 2021-06-18
Encode (escape) TeamCity build link query parameters

### LINKED ISSUE #101: TeamCity build link broken
TeamCity build number is not necessarily a number; for our projects it is a number plus git hash. Yet CustomBuildScanEnhancements.java expects it to be only number and this results in error:
```
org.apache.maven.InternalErrorException: Internal error: com.gradle.scan.plugin.b: Build scan custom link url must be a well-formed http, https, or mailto URL (name: TeamCity build, value: http://bvteamcity02.org.nasdaqomx.com:8111/viewLog.html?buildNumber=11612 [8e26a27]&buildTypeId=Ncsd_NcsdSolution_Core_MonoBuild_Backend).
```
because for this project buildNumber is "11612 [8e26a27]"

May I suggest to use buildId instead of buildNumber? Our TeamCity (v2020.2.2) in its GUI forms links like this:
```https://bvteamcity02.org.nasdaqomx.com/viewLog.html?buildId=5271771v&buildTypeId=Ncsd_NcsdSolution_Core_MonoBuild_Backend```

You could retrieve buildId this way:
```Optional<String> buildId = projectProperty(mavenSession, "teamcity.build.id");```
#104

jthurne:
The build number can contain characters that are not allowed in query strings.

It is not clear if the buildType can also contain disallowed characters, but
there is no harm in encoding it.

Fixes #101 .

### DIFF
--- a/common-custom-user-data-gradle-plugin/src/main/java/com/gradle/CustomBuildScanEnhancements.java
+++ b/common-custom-user-data-gradle-plugin/src/main/java/com/gradle/CustomBuildScanEnhancements.java
@@ -106,7 +106,7 @@ final class CustomBuildScanEnhancements {
                     Properties properties = readPropertiesFile(teamCityConfigFile.get());
                     String teamCityServerUrl = properties.getProperty("teamcity.serverUrl");
                     if (teamCityServerUrl != null) {
-                        String buildUrl = appendIfMissing(teamCityServerUrl, "/") + "viewLog.html?buildNumber=" + buildNumber.get() + "&buildTypeId=" + buildTypeId.get();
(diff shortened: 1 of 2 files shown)


runningcode:
Does this supersede this? https://github.com/gradle/gradle-enterprise-build-config-samples/pull/104
What is better about the build number as opposed to the build id in terms of creating a url?

jthurne:
> Does this supersede this? #104
> What is better about the build number as opposed to the build id in terms of creating a url?

Yes, it does.

We decided to go this route because it's a smaller, less risky change. Additionally, @etiennestuder recalls that the we had found a case in the past where the build ID was not unique.

runningcode (approved):
Makes sense! Thanks.

PR #111, opened 2021-07-06
Use the Git branch name provided by Jenkins when build is run on Jenkins

### LINKED ISSUE #109: Git branch name should be correct on CI
Many CI systems checkout the git project in a detached `HEAD` state leading the branch name to be tagged as `HEAD`.
[Here is an example](https://ge.spring.io/s/vrsotpyyb7xz6#info) from the spring-authorization-server project.

Each CI system has a unique way of passing the branch name as an environment variable. We should try to capture this as best as possible since `HEAD` is not very useful information.

For example on Jenkins, the branch name is set as the environment variable `BRANCH_NAME`.

One proposal is to try to capture the environment variable as a preference, and use the git status as a backup.

let's discuss possible solutions as well as the differences in setting this variable on Ci.

jthurne:
Resolves #109.

### DIFF
--- a/common-custom-user-data-gradle-plugin/src/main/java/com/gradle/CustomBuildScanEnhancements.java
+++ b/common-custom-user-data-gradle-plugin/src/main/java/com/gradle/CustomBuildScanEnhancements.java
@@ -9,6 +9,7 @@ import org.gradle.api.tasks.testing.Test;
 
 import java.util.Optional;
 import java.util.Properties;
+import java.util.function.Supplier;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
@@ -247,8 +248,8 @@ final class CustomBuildScanEnhancements {
             String gitRepo = execAndGetStdOut("git", "config", "--get", "remote.origin.url");
             String gitCommitId = execAndGetStdOut("git", "rev-parse", "--verify", "HEAD");
             String gitCommitShortId = execAndGetStdOut("git", "rev-parse", "--short=8", "--verify", "HEAD");
+            String gitBranchName = getGitBranchName(() -> execAndGetStdOut("git", "rev-parse", "--abbrev-ref", "HEAD"));
             String gitStatus = execAndGetStdOut("git", "status", "--porcelain");
-            String gitBranchName = getGitBranchName();
 
             if (isNotEmpty(gitRepo)) {
                 buildScan.value("Git repository", gitRepo);
@@ -291,14 +292,14 @@ final class CustomBuildScanEnhancements {
             return execAndCheckSuccess("git", "--version");
         }
 
-        private String getGitBranchName() {
+        private String getGitBranchName(Supplier<String> gitCommand) {
             if (isJenkins() || isHudson()) {
                 Optional<String> branch = Utils.envVariable("BRANCH_NAME", providers);
                 if (branch.isPresent()) {
                     return branch.get();
                 }
             }
-            return execAndGetStdOut("git", "rev-parse", "--abbrev-ref", "HEAD");
+            return gitCommand.get();
         }
 
         private boolean isJenkins() {
@@ -308,6 +309,7 @@ final class CustomBuildScanEnhancements {
         private boolean isHudson() {
             return Utils.envVariable("HUDSON_URL", providers).isPresent();
         }
+
     }
 
     private void addCustomValueAndSearchLink(String name, String value) {
--- a/common-custom-user-data-gradle-plugin/src/main/java/com/gradle/CustomBuildScanEnhancements.java
+++ b/common-custom-user-data-gradle-plugin/src/main/java/com/gradle/CustomBuildScanEnhancements.java
@@ -239,8 +239,16 @@ final class CustomBuildScanEnhancements {
             String gitRepo = execAndGetStdOut("git", "config", "--get", "remote.origin.url");
             String gitCommitId = execAndGetStdOut("git", "rev-parse", "--verify", "HEAD");
             String gitCommitShortId = execAndGetStdOut("git", "rev-parse", "--short=8", "--verify", "HEAD");
-            String gitBranchName = execAndGetStdOut("git", "rev-parse", "--abbrev-ref", "HEAD");
             String gitStatus = execAndGetStdOut("git", "status", "--porcelain");
+
+            String gitBranchName;
+            // It is okay to call System.getenv directly here since this lambda is saved to the configuration cache
+            // and re-executed on every build.
+            if (isNotEmpty(System.getenv("BRANCH_NAME"))) {
+                gitBranchName = System.getenv("BRANCH_NAME");
+            } else {
+                gitBranchName = execAndGetStdOut("git", "rev-parse", "--abbrev-ref", "HEAD");
+            }
 
             if (isNotEmpty(gitRepo)) {
                 buildScan.value("Git repository", gitRepo);
--- a/common-custom-user-data-gradle-plugin/src/main/java/com/gradle/CustomBuildScanEnhancements.java
+++ b/common-custom-user-data-gradle-plugin/src/main/java/com/gradle/CustomBuildScanEnhancements.java
@@ -227,10 +227,16 @@ final class CustomBuildScanEnhancements {
     }
 
     private void captureGitMetadata() {
-        buildScan.background(new CaptureGitMetadataAction());
+        buildScan.background(new CaptureGitMetadataAction(providers));
     }
 
-    private static class CaptureGitMetadataAction implements Action<BuildScanExtension> {
+    private static final class CaptureGitMetadataAction implements Action<BuildScanExtension> {
+
+        private final ProviderFactory providers;
+
+        private CaptureGitMetadataAction(ProviderFactory providers) {
+            this.providers = providers;
+        }
 
         @Override
         public void execute(BuildScanExtension buildScan) {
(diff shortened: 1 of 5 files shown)


common-custom-user-data-gradle-plugin/src/main/java/com/gradle/CustomBuildScanEnhancements.java
```diff
@@ -301,7 +309,7 @@ private static void addCustomValueAndSearchLink(BuildScanExtension buildScan, St
         }
     }
 
-    private void captureTestParallelization() {
+    private static void captureTestParallelization(Gradle gradle, BuildScanExtension buildScan) {
```
runningcode, line 312: is this a separate fix for config cache compatibility?

common-custom-user-data-gradle-plugin/src/main/java/com/gradle/CustomBuildScanEnhancements.java
```diff
@@ -239,9 +239,17 @@ private void captureGitMetadata() {
             String gitRepo = execAndGetStdOut("git", "config", "--get", "remote.origin.url");
             String gitCommitId = execAndGetStdOut("git", "rev-parse", "--verify", "HEAD");
             String gitCommitShortId = execAndGetStdOut("git", "rev-parse", "--short=8", "--verify", "HEAD");
-            String gitBranchName = execAndGetStdOut("git", "rev-parse", "--abbrev-ref", "HEAD");
             String gitStatus = execAndGetStdOut("git", "status", "--porcelain");
 
+            String gitBranchName;
+            // It is okay to call System.getenv directly here since this lambda is saved to the configuration cache
+            // and re-executed on every build.
```
runningcode, line 246: interesting. for the standard case where we are calling the git command line, does that mean it will not be re-executed on every build?
i ask because this could mean that the git id changes and the old one will still be used as a custom value.

common-custom-user-data-gradle-plugin/src/main/java/com/gradle/CustomBuildScanEnhancements.java
```diff
@@ -239,9 +239,17 @@ private void captureGitMetadata() {
...
+            if (isNotEmpty(System.getenv("BRANCH_NAME"))) {
```
runningcode, line 247: should we store this as a string on line 243 to be consistent with the other variables?

runningcode (approved):
Looks good. Just one question.

common-custom-user-data-gradle-plugin/src/main/java/com/gradle/CustomBuildScanEnhancements.java
```diff
@@ -239,9 +239,17 @@ private void captureGitMetadata() {
             String gitRepo = execAndGetStdOut("git", "config", "--get", "remote.origin.url");
             String gitCommitId = execAndGetStdOut("git", "rev-parse", "--verify", "HEAD");
             String gitCommitShortId = execAndGetStdOut("git", "rev-parse", "--short=8", "--verify", "HEAD");
-            String gitBranchName = execAndGetStdOut("git", "rev-parse", "--abbrev-ref", "HEAD");
             String gitStatus = execAndGetStdOut("git", "status", "--porcelain");
 
+            String gitBranchName;
```
etiennestuder, line 244: I would replace this with 

`String gitBranchName = ((isJenkins() || isHudson()) && envVariable("BRANCH_NAME).isPresent()) ? envVariable("BRANCH_NAME).get() : execAndGetStdOut("git", "rev-parse", "--abbrev-ref", "HEAD");`

jthurne: I thought about doing that, but then the isJenkins and isHudson functions will need to be made static, and then we'll need to be able to pass in the Providers object. Which we certainly could do, but we will loose out on some of the benefits from the earlier refactoring.

common-custom-user-data-gradle-plugin/src/main/java/com/gradle/CustomBuildScanEnhancements.java
jthurne: The "lambda" is the one that starts on [line 234](https://github.com/gradle/gradle-enterprise-build-config-samples/pull/111/files#diff-e72bae4123cfd74f4e12e8d13bc2c9c5dff46d384cc8f3bbdd6224791c407553R234) and includes the calls to git.  

So the git commands are also called on every build (including the ones that get the git commit IDs).

common-custom-user-data-gradle-plugin/src/main/java/com/gradle/CustomBuildScanEnhancements.java
jthurne: I have some ideas on how to do this and still preserve the benefits of the refactoring. I'll gave them a try and will update the PR accordingly.

common-custom-user-data-gradle-plugin/src/main/java/com/gradle/CustomBuildScanEnhancements.java
jthurne: Yes, although it is in a separate commit from the other changes. I might split it out into it's own PR though so that we can get it merged. I was just being lazy.

common-custom-user-data-gradle-plugin/src/main/java/com/gradle/CustomBuildScanEnhancements.java
jthurne: I just added a commit that attempts to resolve this. It adds a small bit of duplication, but I think it makes a few things clearer/more obvious, while also keeping the diff between the Gradle and Maven versions of the file as similar as possible.

common-custom-user-data-gradle-plugin/src/main/java/com/gradle/CustomBuildScanEnhancements.java
jthurne: I believe the latest commit addresses this.

common-custom-user-data-gradle-plugin/src/main/java/com/gradle/CustomBuildScanEnhancements.java
jthurne: The latest commit converts the Lambda into a full static class which should make this more obvious.

I deleted the comment because I'm worried it might confuse users. I can add a similar comment back in if we think it would be useful.

common-custom-user-data-gradle-plugin/src/main/java/com/gradle/CustomBuildScanEnhancements.java
```diff
@@ -227,20 +227,22 @@ private boolean isBitrise() {
     }
 
     private void captureGitMetadata() {
-        buildScan.background(captureGitMetadataAction);
+        buildScan.background(new CaptureGitMetadataAction());
     }
 
-    private static final Action<BuildScanExtension> captureGitMetadataAction =
-        buildScan -> {
+    private static class CaptureGitMetadataAction implements Action<BuildScanExtension> {
```
etiennestuder, line 233: Can the class be final?

etiennestuder: I would pass Gradle and Providers to this class and store them as fields. This will avoid the null handling you had to introduce when querying env vars. I think...

jthurne: Passing Gradle would break configuration cache compatibility, but luckily we don't need it. And it appears that passing in ProvidersFactory does not break configuration-cache compatibility. So I made that change.

jthurne: Also marked the class as final.

common-custom-user-data-gradle-plugin/src/main/java/com/gradle/Utils.java
```diff
@@ -116,7 +116,7 @@ static Properties readPropertiesFile(String name, ProviderFactory providers, Gra
     }
 
     static InputStream readFile(String name, ProviderFactory providers, Gradle gradle) throws FileNotFoundException {
-        if (isGradle65OrNewer()) {
+        if (isGradle65OrNewer() && providers != null && gradle != null) {
```
etiennestuder, line 119: I believe these null checks can be removed, given your latest changes.

etiennestuder (approved):
I pushed one small refactoring. Please take a look, @jthurne.
Besides that, there is only one comment pending re: the removal of null checks from the Utils class.
This is ready to be merged if you agree with the former and and once you have addressed the latter.

common-custom-user-data-gradle-plugin/src/main/java/com/gradle/CustomBuildScanEnhancements.java
etiennestuder: > And it appears that passing in ProvidersFactory does not break configuration-cache compatibility. 

Note that ProvidersFactory is meant as _the_ intended construct to maintain config cache compatibility. It's not a conincidence storing it as a field works with CC :-)

common-custom-user-data-gradle-plugin/src/main/java/com/gradle/Utils.java
jthurne: Good catch! I had removed the other ones, but I missed these.

jthurne:
> I pushed one small refactoring. Please take a look, @jthurne.

Looks good!

PR #112, opened 2021-07-07
Fix configuration cache bug when capturing Test maxParallelForks

jthurne:
captureTestParallelization needs to be static so that the anonymous inner Action subclass can be serialized without maintaining a reference to the CustomBuildScanEnhancements instance.

### DIFF
--- a/common-custom-user-data-gradle-plugin/CHANGELOG.md
+++ b/common-custom-user-data-gradle-plugin/CHANGELOG.md
@@ -2,6 +2,7 @@
 All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
+- Fix configuration cache compatibility when capturing test maxParallelForks.
 
 ## [1.4.1] - 2021-06-22
 - Encode (escape) the TeamCity build number when creating a link to the TeamCity build (fixes Issue #101).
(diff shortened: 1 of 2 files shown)


runningcode (approved):
Nice find.

etiennestuder:
@jthurne I would have done this as we did it here:
https://github.com/gradle/gradle-enterprise-build-config-samples/blob/c0f24bd0651be4dd245dceca70b1ab3c040e3bcc/common-custom-user-data-gradle-plugin/src/main/java/com/gradle/CustomBuildScanEnhancements.java#L233

I would also have checked if this comment/workaround is still applicable:
https://github.com/gradle/gradle-enterprise-build-config-samples/blob/c0f24bd0651be4dd245dceca70b1ab3c040e3bcc/common-custom-user-data-gradle-plugin/src/main/java/com/gradle/CustomBuildScanEnhancements.java#L308

jthurne:
> @jthurne I would have done this as we did it here:

Ah yes, that's a good idea. I'll make that change and push up a new PR.

One difference in this case is that the BuildScanExtension has to be passed into the closure (since it's an Action<Task> instead of an Action<BuildScanExtension>), so it will be slightly different. But same basic idea.

PR #115, opened 2021-07-20
Use the new GradleEnterpriseListener interface

facewindu:
This PR changes quite a bit the structure of the CCUDME project.
It is now an EventSpy that register very early (in its init callback) a custom listener implementing the new `GradleEnterpriseListener` interface. This allows us to get rid of Maven callbacks quirks and totally control what's going on here (instead of e.g. relying on AbstractMavenLifecycleParticipant callbacks being fired 'at the right time')


common-custom-user-data-maven-extension/src/main/java/com/gradle/maven/common/user/data/extension/ApiAccessor.java
```diff
@@ -0,0 +1,72 @@
...
+import org.apache.maven.MavenExecutionException;
+import org.codehaus.plexus.PlexusContainer;
+import org.codehaus.plexus.classworlds.realm.ClassRealm;
+import org.codehaus.plexus.logging.Logger;
+
+import java.util.Optional;
+
+import static java.util.Comparator.comparing;
+
+final class ApiAccessor {
```
guylabs, line 12: Are we able to provide this via the Plexus container from the Maven extension itself such that this code is handled on our side and then injected into the custom event spy?

common-custom-user-data-maven-extension/src/main/java/com/gradle/maven/common/user/data/extension/CommonCustomUserDataMavenExtensionEventSpy.java
```diff
@@ -0,0 +1,37 @@
...
+
+import org.apache.maven.eventspy.AbstractEventSpy;
+import org.apache.maven.eventspy.EventSpy;
+import org.codehaus.plexus.PlexusContainer;
+import org.codehaus.plexus.component.annotations.Component;
+import org.codehaus.plexus.logging.Logger;
+
+import javax.inject.Inject;
+
+@Component(
```
guylabs, line 11: Is this event spy required ? Can't we call `ApiAccessor.registerExtension(getClass(), container, logger);` inside the constructor of the `CommonCustomUserDataGradleEnterpriseListener` where we do have access to the `PlexusContainer` as well?

common-custom-user-data-maven-extension/src/main/java/com/gradle/maven/common/user/data/extension/ApiAccessor.java
facewindu: no, because this code is especially about sharing things from the Maven extension to other places.

The only way to do that would be to hack parent realms from the maven extension side.
Basically, store in 'plexus.core' or another realm that is visible to all extensions in all cases the things we want to be visible to all extensions.
That would be possible I guess (it's another version of Marc's hack) but really not what Maven design wants us to do. (Marc's hack is about [this bug](https://issues.apache.org/jira/browse/MNG-6906) but does not violate Maven's principles, because if we understood exported packages properly, it should have worked even without the hack in all cases)

Then, even if we were ok doing this, we would still have to do the other way around: make the `GradleEnterpriseListener` implementation on the consuming extension available on the Maven extension side:
(this guy:
```
importPackage(extensionRealm, maybeGERealm.get(), OWN_PACKAGE);
```
)

common-custom-user-data-maven-extension/src/main/java/com/gradle/maven/common/user/data/extension/CommonCustomUserDataMavenExtensionEventSpy.java
facewindu: There are cases where the `CommonCustomUserDataGradleEnterpriseListener` is not constructed, even if marked with the `@Component` annotation.
But even there, it needs the `GradleEnterpriseListener` interface at runtime when constructed, so the interface needs to be available in its realm when constructed.(and luckily, it's constructed after `EventSpy#init` is called)

common-custom-user-data-maven-extension/src/main/java/com/gradle/maven/common/user/data/extension/CustomBuildScanEnhancements.java
```diff
@@ -1,35 +1,37 @@
-package com.gradle;
+package com.gradle.maven.common.user.data.extension;
```
marcphilipp, line 1: Maybe a bit shorter?
```suggestion
package com.gradle.maven.custom.extension;
```

common-custom-user-data-maven-extension/src/main/java/com/gradle/maven/common/user/data/extension/CommonCustomUserDataGradleEnterpriseListener.java
```diff
@@ -1,60 +1,58 @@
...
+import javax.annotation.Nullable;
 import javax.inject.Inject;
 
 @Component(
-    role = AbstractMavenLifecycleParticipant.class,
-    hint = "common-custom-user-data",
-    description = "Captures common custom user data in Maven build scans"
+        role = GradleEnterpriseListener.class,
+        hint = "common-custom-user-data",
+        description = "Captures common custom user data in Maven build scans"
```
marcphilipp, line 18: Is this how IDEA indents it?

common-custom-user-data-maven-extension/src/main/java/com/gradle/maven/common/user/data/extension/CommonCustomUserDataGradleEnterpriseListener.java
```diff
@@ -1,60 +1,58 @@
...
             logger.debug("Finished configuring build cache");
 
-            GroovyScriptUserData.evaluate(session, gradleEnterprise, logger);
+            GroovyScriptUserData.evaluate(session, api, logger);
             SystemPropertyOverrides.configureBuildCache(buildCache);
-        } else {
-            logger.debug("Could not find GradleEnterpriseApi component in Plexus container");
+        } catch (MavenExecutionException e) {
+            logger.error("Error executing Common Custom User Data Maven extension:");
+            logger.error(e.getMessage(), e);
```
marcphilipp, line 55: Should we maybe handle this at the call site in the GE Maven extension?

common-custom-user-data-maven-extension/src/main/java/com/gradle/maven/common/user/data/extension/GroovyScriptUserData.java
```diff
@@ -8,11 +8,16 @@
...
 
+import javax.annotation.Nullable;
 import java.io.File;
 
 final class GroovyScriptUserData {
 
-    static void evaluate(MavenSession session, GradleEnterpriseApi gradleEnterprise, Logger logger) throws MavenExecutionException {
+    static void evaluate(@Nullable MavenSession session, GradleEnterpriseApi gradleEnterprise, Logger logger) throws MavenExecutionException {
+        if (session == null) {
+            logger.warn("Cannot evaluate groovy script when the Maven session in not available");
```
marcphilipp, line 18: Should we instead document `session` and `project` as nullable and evaluate it anyway?

common-custom-user-data-maven-extension/pom.xml
```diff
@@ -65,7 +65,7 @@
         <dependency>
             <groupId>com.gradle</groupId>
             <artifactId>gradle-enterprise-maven-extension</artifactId>
-            <version>1.10.3</version>
+            <version>1.11</version>
```
marcphilipp, line 68: Will we have to wait for the 1.11 release before merging this PR?

etiennestuder:
I want to review this PR before it gets merged.

etiennestuder:
@facewindu why did the package names have to change?

facewindu:
> I want to review this PR before it gets merged.

don't worry @etiennestuder you will. I wanted to review it between Guy, Marc and me before asking the solutions team. It will need the Maven extension 1.11 anyway, as a dependency.

common-custom-user-data-maven-extension/src/main/java/com/gradle/maven/common/user/data/extension/ApiAccessor.java
facewindu: Spent many hours on this, and finally got something. Which made the consuming part (i.e. the CCUDME) much nicer and shorter.

common-custom-user-data-maven-extension/src/main/java/com/gradle/maven/common/user/data/extension/CustomBuildScanEnhancements.java
facewindu: I went back with the existing package  `com.gradle`. We originally needed to change it, because I was importing it back to the GE extension realm, and having sth not precise enough was causing some recursion in class loading. This is not a problem anymore as we don't import anything back

common-custom-user-data-maven-extension/src/main/java/com/gradle/maven/common/user/data/extension/CommonCustomUserDataGradleEnterpriseListener.java
facewindu: for me, yes

common-custom-user-data-maven-extension/src/main/java/com/gradle/maven/common/user/data/extension/GroovyScriptUserData.java
facewindu: not sure what you mean. If the session is not available (very edge case scenarios with very early Maven build failures), we cannot grab the location of the groovy script (from `session.getRequest().getMultiModuleProjectDirectory()` +`.mvn/gradle-enterprise-custom-user-data.groovy`)

common-custom-user-data-maven-extension/pom.xml
facewindu: yes, we will have to wait for BA Maven 1.11 to be out first. This version will ship the new `GradleEnterpriseListener` interface, which is used instead of hooking ourselves into Maven lifecycle directly (via `AbstractMavenLifecycleParticipant` or event spies)

facewindu:
> @facewindu why did the package names have to change?

it's not needed anymore

facewindu:
@guylabs @marcphilipp please review again

@gradle/ge-solutions please review this PR.
This builds on top of https://github.com/gradle/dotcom/pull/10070 which allows to hook directly into our own lifecycle, not relying on Maven's.
This has the benefit to really control _when_ the callback is called, and ensure it is really always called. (It was not in case of very early Maven failures otherwise)

We also don't need to hacks to import the GE Maven extension API classes into this extension's realm anymore. It's all done on the GE Maven extension side, by inspecting existing class realms (and realms added after the GE Maven extension is initialized).

Basically now, the CCUDME only provides a `@Component` which implements our own public `GradleEnterpriseListener` interface. Everything is much simple on the consuming side, and more reliable.

I bumped the CCUDME version to 1.8 (new minor) because of this rather impactful change.
It also requires GE Maven extension 1.11+ (no backward compatibility, but I understood this was not an issue)

common-custom-user-data-maven-extension/src/main/java/com/gradle/maven/common/user/data/extension/CommonCustomUserDataGradleEnterpriseListener.java
facewindu: yes, I think this is a better idea. If the others are ok, I will do this.
Basically `GradleEnterpriseListener#configure` will declare that it can throw a `MavenExecutionException`, and we'll catch it around each listener call on the GE Maven extension side.

common-custom-user-data-maven-extension/README.md
```diff
@@ -24,7 +24,8 @@ This table details the version compatibility of the Common Custom User Data Mave
 
 | Common Custom User Data Maven extension versions | Gradle Enterprise Maven extension versions |
 | ------------------------------------------------ | ------------------------------------------ |
-| `1.7+`                                            | `1.10.1+`                                  |
+| `1.8+`                                           | `1.11+`                                  |
```
runningcode, line 27: What type of error message do we get if the versions are incompatible? Will it simply not do anything and silently fail?

runningcode (approved):
Looks good to me. I'm no maven expert though. @bigdaz might have some thoughts.

common-custom-user-data-maven-extension/pom.xml
```diff
@@ -65,7 +65,7 @@
         <dependency>
             <groupId>com.gradle</groupId>
             <artifactId>gradle-enterprise-maven-extension</artifactId>
-            <version>1.10.4</version>
+            <version>1.11-branch-francois_maven_test-prerelease</version>
```
bigdaz, line 68: Since this is used as a template for customers, we should avoid having unreleased versions merged into `master`. I guess we can keep this PR in a branch until `1.11` is released.

common-custom-user-data-maven-extension/src/main/java/com/gradle/CommonCustomUserDataGradleEnterpriseListener.java
```diff
@@ -1,60 +1,54 @@
...
+    @Requirement
+    private Logger logger;
 
     @Override
-    public void afterProjectsRead(MavenSession session) throws MavenExecutionException {
-        logger.debug("Executing extension: " + getClass().getSimpleName());
-
-        GradleEnterpriseApi gradleEnterprise = ApiAccessor.lookupGradleEnterpriseApi(container, getClass());
-        if (gradleEnterprise != null) {
+    public void configure(GradleEnterpriseApi api, @Nullable MavenSession session) {
```
bigdaz, line 27: Is there not a way to implement the caller of `configure` such that `MavenSession` is never null? 
Forcing the client to deal with this seems unfortunate.

bigdaz (approved):
I think this is a big improvement, making the CCUDME a first-class collaborator with the GE Maven extension.
But we should not merge this into `master` until GE Maven extension 1.11 is released. Otherwise customers will not be able to use these sources as a template for their own custom extension.

common-custom-user-data-maven-extension/pom.xml
facewindu: yes, that was the plan. I'll change the PR to contain 1.11 when released and we will merge it after. Keeping it like that so that you can play around with the custom 1.11 for now

common-custom-user-data-maven-extension/src/main/java/com/gradle/CommonCustomUserDataGradleEnterpriseListener.java
facewindu: I just made changes on https://github.com/gradle/dotcom/pull/10070 and here, as a consequence, to do just that. It's indeed better.

common-custom-user-data-maven-extension/src/main/java/com/gradle/maven/common/user/data/extension/CommonCustomUserDataGradleEnterpriseListener.java
facewindu: done

common-custom-user-data-maven-extension/src/main/java/com/gradle/maven/common/user/data/extension/GroovyScriptUserData.java
facewindu: moot

common-custom-user-data-maven-extension/README.md
facewindu: It's silently ignored. We do get a 
```
TRACE: Sisu - Ignoring Plexus role: com.gradle.maven.extension.api.GradleEnterpriseListener
java.lang.TypeNotPresentException: Type com.gradle.maven.extension.api.GradleEnterpriseListener not present
...
```
exception logged at the `trace` level (the deepest) but no one will see that.

Given the choice I made to _only_ provide a `@Component` in the extension, we cannot do more than that. We are sure the components will be registered on 'SessionStarted' for instance, so we could also provide a `AbstractMavenLifecycleParticipant` that checks if the `GradleEnterpriseListener` type is present on the CCUDME realm, and log sth if that's not the case, but I think this is taking a step back (complexifying back the CCUDME).

guylabs (approved):

guylabs:
@etiennestuder can you review this as well such that we can finalize it once @facewindu is back next week? Thanks.

marcphilipp (approved):

facewindu:
This is fully ready, including declaring the yet-to-be-released Maven extension 1.11 in the pom.
@etiennestuder please have a last look.
This PR cannot be merged before Maven extension 1.11 is released

etiennestuder (approved):
Wonderful!

PR #128, opened 2021-07-21
Update dependency versions for common-custom-user-data-maven-extension

bigdaz:
This combines 5 updates suggested by dependabot.
There appears to be no change in behaviour in the extension.

### DIFF
--- a/common-custom-user-data-maven-extension/pom.xml
+++ b/common-custom-user-data-maven-extension/pom.xml
@@ -144,7 +144,7 @@
                     <plugin>
                         <groupId>org.apache.maven.plugins</groupId>
                         <artifactId>maven-release-plugin</artifactId>
-                        <version>3.0.0-M1</version>
+                        <version>3.0.0-M4</version>
                         <configuration>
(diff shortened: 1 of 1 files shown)


bigdaz:
I'm going to optimistically merge this, since I've tested this and the extension displays no change in behaviour.
The `plexus` upgrade will mean we no longer work with Maven 2, but that's not supported anyway.

etiennestuder:
Please add a note to the changelog.

bigdaz:
@etiennestuder Done: https://github.com/gradle/gradle-enterprise-build-config-samples/blob/master/common-custom-user-data-maven-extension/CHANGELOG.md

Although I'm not sure this is really necessary. I checked and the binary generated is identical, except for the POM file that is packaged in the jar. I guess this is because:
- `plexus-component-annotations` is a `provided` dependency, and so is not included in our shaded jar
- The changes to `groovy` are likely not used and thus removed from the shaded jar

etiennestuder:
@bigdaz Ah. I see. I did not anticipate that his might not affect the shipped artifact. Thanks.

PR #140, opened 2021-08-24
Add .gitattributes to enforce crlf on .bat files

alextu:


runningcode (approved):

etiennestuder (approved):
LGTM. There is more to it than I had thought: https://www.git-scm.com/docs/gitattributes

PR #160, opened 2021-09-29
Sanitize 'Git repository' custom value

gabrielfeo:
Remove user info from the git remote URL custom value.

URIs can contain username and password, and it's not uncommon for Git remote URLs to use this for HTTP authentication. Users of the `common-custom-user-data` plugins might have their credentials leaked with the "GIt repository" custom value sent in build scans.


gabrielfeo:
Hello there 👋🏻 
Any feedback is much appreciated! If this looks good, I'd like to make the same change in the maven extensions

etiennestuder:
Thanks for the usggestion. I'll provide feedback. But it will take a moment to get on it.

etiennestuder:
Proposal to @bigdaz was to show `***@` instead of just blanking out the user info. This makes it easier to understand in build scans that there were credentials passed as part of the Git url.

bigdaz:
@etiennestuder I'm happy with that proposal, although I wonder if we really want to show the Git Repository, not the Git URL used to access the repository. In that case, we should normalize these 2 to the same value:

`https://github.com/gradle/gradle-enterprise-build-config-samples.git`
`git@github.com:gradle/gradle-enterprise-build-config-samples.git`

etiennestuder:
> @etiennestuder I'm happy with that proposal, although I wonder whether it would be better to show the actual Git Repository, not the Git URL used to access the repository. If we did that, we would want to normalize these 2 to the same value:
> 
> `https://github.com/gradle/gradle-enterprise-build-config-samples.git` `git@github.com:gradle/gradle-enterprise-build-config-samples.git`
> 
> But for sure, that's a separate step from removing sensitive user information from the URL.

I propose we keep the scope of this PR on just removing the user sensitive info and we consider the normalisation you proposed in a separate PR in the future.

bigdaz:
@gabrielfeo Thanks for the contribution. I used this as inspiration for #196, which subsumes this PR.

A few differences:
- I merged the 2 utility methods into one
- Change is applied to Maven extension as well as Gradle plugin
- I avoided introducing test infrastructure for this change. The absence of tests in this repo was a deliberate decision to make the code easier to fork/adapt as an example.

gabrielfeo:
@bigdaz @etiennestuder Thank you, those are good changes. Happy to have helped.

PR #163, opened 2021-10-11
Add GitHub actions to assess CCUD build and samples syntax

jprinet:
- signArchives task is excluded from CCUD smoke test to avoid passing credentials
- Gradle configuration samples are validated on the syntax aspect only

### DIFF
--- a/.github/workflows/gradle-smoke-test.yml
+++ b/.github/workflows/gradle-smoke-test.yml
@@ -24,18 +24,12 @@ jobs:
         strategy:
             matrix:
                 include:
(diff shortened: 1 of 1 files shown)


ghale:
> I agree, but let's limit the scope of our first CI integration. :) We can add that later.

That's exactly what I meant.  This is a good first step.

runningcode:
Regarding renaming - I think we should use the name `gradle-slow-task-executions.gradle` since it refers to tasks. Not all work units are not captured here.

PR #165, opened 2021-10-13
Update `log` binding for renamed type in Maven nightly

bigdaz:
We previously used `org.apache.maven.monitor.logging.DefaultLog` to wrap
a Plexus `Logger` instance as a `org.apache.maven.plugin.logging.Log` type.
This type has been renamed in the latest Maven nightly, so we now provide
a custom wrapper implementation instead.

This is a quick-fix to ensure the CCUDME functions with all supported versions
of Maven. A better solution might be to provide a SLF4J logger to Groovy
scripts rather than `org.apache.maven.plugin.logging.Log`.


etiennestuder:
@bigdaz Why don't we pass the raw org.codehaus.plexus.logging.Logger to the groovy script, instead of wrapping it with something? (a question also appliccable to the code before your PR)

Issue #164, opened 2021-10-12
Use SLF4J logger instead of DefaultLog


guylabs:
In the latest Maven nightly version the `DefaultLog` class [has been renamed](https://github.com/apache/maven/commit/c3962c1a6c0b40abbfb12740253437f6431b234b#diff-08e7da438372645395143b0c54c0b672b7a63abb27b2539e012bf0770edc9ff5).

This makes [nightly tests fail](https://builds.gradle.org/buildConfiguration/Enterprise_Release_Component_CommonCustomUserDataMavenExtension_CrossVersionNightly_MavenLatestNightlyLinux?mode=builds#all-projects) as we use the `DefaultLog` class here https://github.com/gradle/gradle-enterprise-build-config-samples/blob/d389428132f95a713eaaafdf2c6d43c81bde84fb/common-custom-user-data-maven-extension/src/main/java/com/gradle/GroovyScriptUserData.java#L46

We propose to use the SLF4J logger instead as this is stable, but this would mean that some users might have issues as this is a breaking change as the API of the `DefaultLog` is different.

etiennestuder:
Can you please also ask @marcphilipp for review. See also my question [here](https://github.com/gradle/gradle-enterprise-build-config-samples/pull/165#issuecomment-941918289).

marcphilipp:
I'm also in favor of #166.

PR #166, opened 2021-10-13
Provide a SLF4J logger to CCUDME Groovy scripts

bigdaz:
When a groovy script is used to add custom tags/links/values via the
common-custom-user-data-maven-extension, a `log` variable is bound to allow
custom messages to be logged.

Previously this was of type `org.apache.maven.plugin.logging.Log`, using a `DefaultLog`
instance to wrap a Plexus Logger. With this change, we provide a standard `org.slf4j.Logger`
instance to Groovy scripts instead, reducing our dependency on core Maven types.

This is a minor breaking change as the `org.slf4j.Logger` type differs slightly from
`org.apache.maven.plugin.logging.Log`, and some groovy scripts will no longer function.

### DIFF
--- a/common-custom-user-data-maven-extension/CHANGELOG.md
+++ b/common-custom-user-data-maven-extension/CHANGELOG.md
@@ -2,6 +2,7 @@
 All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
+- Change the type of the `log` variable bound for custom Groovy scripts to `org.slf4j.Logger`. This change may require changes in custom Groovy scripts.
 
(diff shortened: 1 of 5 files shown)


bigdaz:
> I think we should indicate this as a potentially breaking change by incrementing the major version number.

I've pushed a change to bump the version to `2.0`. While this is strictly correct, I'm not sure it's not overkill in this instance.
See here for more details: https://github.com/gradle/gradle-enterprise-build-config-samples/issues/164#issuecomment-942686534

PR #188, opened 2021-11-02
Apply build scan enhancements after settings are evaluated

### LINKED ISSUE #187: Custom value search links are not generated with CCUD Gradle plugin
When running a build on CI, we attempt to generate custom values and associated search links to list other build scans with the same custom value. For example, we may create a `CI Job` custom value and a `CI Job build scans` custom link.

In most scenarios, the custom value is generated, but the custom link is not generated. This is because at the time the CCUD Gradle plugin is being applied, the `gradleEnterprise.buildScan.server` value has not yet been set. The CCUD plugin uses `settings.getPluginManager().withPlugin("com.gradle.enterprise", __ -> {` to execute the custom rules, which will execute before the `gradleEnteprise` extension has been configured.

An exception to this is the `Git commit id build scans` link, which is successfully generated (usually) since the code is run in a `buildScan.background` block. Note that in some instances I've seen this link not generated as well.

A workaround to this issue is to apply the CCUD plugin after the `gradleEnterprise` extension has been configured:
```
plugins {
    id 'com.gradle.enterprise' version '3.7.1'
    id 'com.gradle.common-custom-user-data-gradle-plugin' version '1.4.2' apply false
}

gradleEnterprise {
    server = "https://e.grdev.net"
}

apply plugin: 'com.gradle.common-custom-user-data-gradle-plugin'
```

bigdaz:
Previously, the build scan enhancements were applied immediately on plugin application.
This meant that custom-link rules that required the GE server URL could not be
run, as the server property had not yet been set.

With this change, all build scan enhancements are run after the settings has been
evaluated. At this stage, the gradleEnterprise extension should be fully configured.

Fixes #187

### DIFF
--- a/common-custom-user-data-gradle-plugin/src/main/java/com/gradle/CommonCustomUserDataGradlePlugin.java
+++ b/common-custom-user-data-gradle-plugin/src/main/java/com/gradle/CommonCustomUserDataGradlePlugin.java
@@ -42,10 +42,6 @@ public class CommonCustomUserDataGradlePlugin implements Plugin<Object> {
             GradleEnterpriseExtension gradleEnterprise = settings.getExtensions().getByType(GradleEnterpriseExtension.class);
             customGradleEnterpriseConfig.configureGradleEnterprise(gradleEnterprise);
 
-            BuildScanExtension buildScan = gradleEnterprise.getBuildScan();
-            customGradleEnterpriseConfig.configureBuildScanPublishing(buildScan);
-            new CustomBuildScanEnhancements(buildScan, providers, settings.getGradle()).apply();
-
             BuildCacheConfiguration buildCache = settings.getBuildCache();
             customGradleEnterpriseConfig.configureBuildCache(buildCache);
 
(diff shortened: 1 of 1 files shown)


runningcode (approved):
Does this mean that in the currently release version we are missing all these build scan links?

alextu:
According to the itests I've started, most of the time `buildScan.getServer()` is null (so no links) but it might just be pure (un)luck. I also confirm that Daz change fixes it (just checked running my itests).

alextu (approved):

bigdaz:
> Does this mean that in the currently release version we are missing all these build scan links?

@runningcode Yes, I think this is indeed the case.

etiennestuder:
A problem with this change is that whenver something goes during the setting evaluation, there will be no tags at all on the build scan. Moving the entire build scans enhancements logic to afterEvaluate seems to broad to me - I would only do the link creation in the callback.

etiennestuder (commented):
Trying to provide the same feedback as part of the GH review workflow:

A problem with this change is that whenver something goes wrong during the settings evaluation, there will be no tags at all on the build scan. Moving the entire build scans enhancements logic to afterEvaluate seems too broad to me - I would only do the link creation in the callback.

etiennestuder:
I would do the following: instead of directly adding the custom links, add them to a list, and then trigger the processing of the list from the outside. this way, it will work with >6, and <6 Gradle versions.

bigdaz:
Reverted and improved by #192

PR #191, opened 2021-11-04
Add tag when Intellij Idea sync build is detected

alextu:
We should provide a specific tag on build scans resulting from an IDEA sync (when hitting the "Reload All Gradle Projects" button)

### DIFF
--- a/common-custom-user-data-gradle-plugin/src/main/java/com/gradle/CustomBuildScanEnhancements.java
+++ b/common-custom-user-data-gradle-plugin/src/main/java/com/gradle/CustomBuildScanEnhancements.java
@@ -57,7 +57,10 @@ final class CustomBuildScanEnhancements {
                 Optional<String> newIdeaVersion = sysProperty("idea.version");
                 Optional<String> oldIdeaVersion = firstSysPropertyKeyStartingWith("idea.version");
                 Optional<String> eclipseVersion = sysProperty("eclipse.buildId");
-
+                Optional<String> ideaSync = sysProperty("idea.sync.active");
+                if (ideaSync.isPresent()) {
+                    buildScan.tag("IDE sync");
+                }
                 if (invokedFromAndroidStudio.isPresent()) {
                     buildScan.tag("Android Studio");
                     androidStudioVersion.ifPresent(v -> buildScan.value("Android Studio version", v));


alextu:
I don't remember having seen such similar "sync" builds in Intellij's maven integration, have you?

bigdaz:
AFAIK, Intellij will not invoke Maven in order to build the IDE model for a Maven project. If that's true, then there won't be any event for us to hook into.

bigdaz:
I think it's worth reviewing all of the IDEA / Android Studio tags/values. 

Reason: I was curious why we were detecting a _project property_ when [the only reference I could find to `idea.sync.active`](https://stackoverflow.com/questions/63409895/configure-gradle-plugin-based-on-future-tasks) mentions a _system property_. So I decided to try things out for myself, and when importing a vanilla gradle project into IntelliJ with CCUDGP 1.4.2, I got `Android Studio` tag!

Build scan here: https://scans.gradle.com/s/qlww2tmxrsdau

For reference, I put together a simple project that prints all of the properties starting with 'idea' or 'android':
IDEA sync: https://scans.gradle.com/s/qlww2tmxrsdau/console-log?anchor=29
Android Studio sync: https://scans.gradle.com/s/ge5rshkw7kpuq/console-log?anchor=34
Run 'clean' from IDEA: https://scans.gradle.com/s/gkqkr74vd5lhm/console-log?anchor=40
Run 'clean' from Android Studio: https://scans.gradle.com/s/dfk2bgugf2dia/console-log?anchor=31

It looks like we could be using `idea.platform.prefix` to tell between IDEA and Android Studio, and use `idea.sync.active` (system property) to determine if it's a sync or not. Sync can apply to both IDEA and AS. However, we might want to check with older versions of these tools too, and also check if these work for IDEA when Android Studio is _not_ installed.

bigdaz (changes requested):
See my main comment on the PR. I think we should rework/retest all of the detection of IDEA and Android Studio.

alextu:
- The sync tag:
Thanks Daz, agreed I tested the system property at the time and then reverted the changes and then re-coded it quickly yesterday but with a project property without manually retesting... 🤦‍♂️ 
Yes it's relevant for both Intellij and Android Studio, do you suggest a more generic name for the tag?

- IDEA vs Android Studio detection
Interesting a month ago I asked the same question to Jetbrains https://youtrack.jetbrains.com/issue/IDEA-279187#focus=Comments-27-5233519.0-0 and they could not give me a reliable property to distinguish between IDEA and Android Studio.
So I tried on my side and indeed the `idea.platform.prefix` does not exist anymore for Intellij Idea `2021.2.3` 🤷.
I think in current state of things we should not try to distinguish between Android Studio and Intellij and just tag with `IDEA` or something broad enough.

bigdaz:
> The sync tag: do you suggest a more generic name for the tag?

I think a shared tag name makes sense for Android studio and IntelliJ. Perhaps `IDE sync` would be better? (We could possibly add support for Eclipse sync as well, and use the same tag).

> I think in current state of things we should not try to distinguish between Android Studio and Intellij and just tag with IDEA or something broad enough.

I would only do this as a last resort. First I think we could collect the properties for different AS and IDEA versions, and come up with something that works in most cases, preferring 'IDEA' over 'Android Studio' when we aren't certain.

bigdaz:
> I would only do this as a last resort. First I think we could collect the properties for different AS and IDEA versions, and come up with something that works in most cases, preferring 'IDEA' over 'Android Studio' when we aren't certain.

@alextu I suggest splitting this into a separate issue. The fact that we display "Android Studio" for an IntelliJ sync is a clear bug in my mind. (Android Studio _is a variant of_ IntelliJ IDEA, but not vice versa).

alextu:
> I think a shared tag name makes sense for Android studio and IntelliJ. Perhaps IDE sync would be better? (We could possibly add support for Eclipse sync as well, and use the same tag).

Sounds good, will change it to `IDE sync`

> 
@alextu I suggest splitting this into a separate issue. The fact that we display "Android Studio" for an IntelliJ sync is a clear bug in my mind. (Android Studio is a variant of IntelliJ IDEA, but not vice versa).

I agree that it should be a separate issue that can be discussed further there.

etiennestuder:
I suggest we pause this PR for a moment, integrate the other PRs, and do a release - and come back to this one a bit later.

etiennestuder (approved):

PR #192, opened 2021-11-04
Second attempt at fixing custom link generation

bigdaz:
This is another attempt at fixing #187. Instead of deferring all build scan customization, we only defer the creation of custom links that require the server url to be configured.

The final commit in this PR also defers the capture of Git commit metadata, avoid a potential (but rare) race condition.
This is not perfect, and we could either:
- Ignore this potential race condition, and not change the capture of Git commit metadata
- Implement a thread-safe solution to defer only the generation of the "git commit id build scans" link

### DIFF
--- a/common-custom-user-data-gradle-plugin/src/main/java/com/gradle/CommonCustomUserDataGradlePlugin.java
+++ b/common-custom-user-data-gradle-plugin/src/main/java/com/gradle/CommonCustomUserDataGradlePlugin.java
@@ -44,7 +44,8 @@ public class CommonCustomUserDataGradlePlugin implements Plugin<Object> {
 
             BuildScanExtension buildScan = gradleEnterprise.getBuildScan();
             customGradleEnterpriseConfig.configureBuildScanPublishing(buildScan);
-            new CustomBuildScanEnhancements(buildScan, providers, settings.getGradle()).apply();
+            CustomBuildScanEnhancements buildScanEnhancements = new CustomBuildScanEnhancements(buildScan, providers, settings.getGradle());
+            buildScanEnhancements.apply();
 
             BuildCacheConfiguration buildCache = settings.getBuildCache();
             customGradleEnterpriseConfig.configureBuildCache(buildCache);
@@ -55,6 +56,8 @@ public class CommonCustomUserDataGradlePlugin implements Plugin<Object> {
                 SystemPropertyOverrides overrides = new SystemPropertyOverrides(providers);
                 overrides.configureGradleEnterprise(gradleEnterprise);
                 overrides.configureBuildCache(buildCache);
+
+                buildScanEnhancements.runPostEvaluateActions();
             });
         });
     }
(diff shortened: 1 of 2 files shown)


etiennestuder:
Thanks, @bigdaz . I'll take a look today.

etiennestuder (approved):
LGTM. Thanks for the refactoring.

etiennestuder:
@bigdaz Did you consder whether there is also something on the CCUD Maven plugin that could/should be improved similar to what you did here for Gradle?

etiennestuder:
@bigdaz I leave deleting the branch to you.

bigdaz:
> @bigdaz Did you consider whether there is also something on the CCUD Maven plugin that could/should be improved similar to what you did here for Gradle?
> 
> I think we have the same issue on the Maven side, e.g. if the server prop is overridden via sys props, the custom links would currently point to the wrong server.

I didn't test this, because I assumed that all of the GE configuration (including property overrides) would be handled when loading the GE Maven extension, before the CCUD Maven extension was loaded.
Looking at the code now, the only System property override that is applied by the CCUD extension is `gradle.cache.remote.shard`: the `gradle.enterprise.url` property is handled by the GE Maven extension.

I didn't really consider that the Maven extension could be doing some configuration of the GE model _after_ the CCUD extension is loaded. I'll do some testing to see if this is possible.

etiennestuder:
I think (@marcphilipp or @facewindu will know for sure) that the sys props understood by the GE Maven plugin will be applied only after our listener in the custom extension is invoked. Thus, all custom links pointing back to GE that we add might be wrong if the GE server url is defined via sys prop. I created a separate issue to look into this at some point (or now, depending on availability).

bigdaz:
> For the Maven case: The `GradleEnterpriseListener#configure` callbacks (so the CCUDME implementation) are executed just before the API is locked. There is a little chance that sth modifies the GA API properties if there is another `GradleEnterpriseListener` implementation that is fired after you.

My testing shows this to be the case. Specifying `-Dgradle.enterprise.url=https://localhost` on the command line causes the correct search links to be generated, even if the value in `gradle-enterprise.xml` is missing or incorrect.

> For the Gradle case:
> You could do your stuff in a buildScan#buildFinished block (which is executed just before we lock the API)

I think this is a better fix that doing it in `settingsEvaluated` or `afterEvaluate`. I'll submit a separate PR with this change.

bigdaz:
PR to use `buildFinished` for custom link generation: #197

facewindu:
> > For the Maven case: The `GradleEnterpriseListener#configure` callbacks (so the CCUDME implementation) are executed just before the API is locked. There is a little chance that sth modifies the GA API properties if there is another `GradleEnterpriseListener` implementation that is fired after you.
> 
> My testing shows this to be the case. Specifying `-Dgradle.enterprise.url=https://localhost` on the command line causes the correct search links to be generated, even if the value in `gradle-enterprise.xml` is missing or incorrect.

The system properties are initialized after session start, and they take precedence over XML configuration.
But you can still avec two implementation of the `GradleEnterpriseListener` interface.
Implementation 1 setting your links / custom values based on the GE server url
Implementation 2 modifying the GE server url, invalidating your captured data

We need a callback after locking for this kind of things, if we want to ensure that the GE server url (amongst other things) is finalized.

> 
> > For the Gradle case:
> > You could do your stuff in a buildScan#buildFinished block (which is executed just before we lock the API)
> 
> I think this is a better fix that doing it in `settingsEvaluated` or `afterEvaluate`. I'll submit a separate PR with this change.

It's better but still suffers from the same fundamental problem.

PR #196, opened 2021-11-08
Redact any user:password contained in Git repo URL

bigdaz:
This user information should not be included as part of the 'Git Repository'
custom value. With this change, any `user:password@` part of the URL will
be replaced by `******@`.

This change was based on PR #160.

Normalizing all 'Git Repository' values is not part of this fix. A later change
could potentially normalize all Git connection URLs to the canonical Git Repository URL.

### DIFF
--- a/common-custom-user-data-gradle-plugin/CHANGELOG.md
+++ b/common-custom-user-data-gradle-plugin/CHANGELOG.md
@@ -3,7 +3,7 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 - Add custom value and search link for GitHub Actions run ID.
-- Fix generation of search links for custom values (#187).
+- Fix generation of search links for custom values.
 - Redact any `user:password@` portion of the 'Git repository' custom value.
 
 ## [1.4.2] - 2021-07-13
--- a/common-custom-user-data-gradle-plugin/CHANGELOG.md
+++ b/common-custom-user-data-gradle-plugin/CHANGELOG.md
@@ -4,6 +4,7 @@ All notable changes to this project will be documented in this file.
 ## [Unreleased]
 - Add custom value and search link for GitHub Actions run ID.
 - Fix generation of search links for custom values (#187).
+- Redact any `user:password@` portion of the 'Git repository' custom value.
 
 ## [1.4.2] - 2021-07-13
 - Fix configuration cache compatibility when capturing test maxParallelForks.
--- a/common-custom-user-data-gradle-plugin/src/main/java/com/gradle/Utils.java
+++ b/common-custom-user-data-gradle-plugin/src/main/java/com/gradle/Utils.java
@@ -106,14 +106,12 @@ final class Utils {
         }
     }
 
-    /**
-     * Redact any `user:password` that is part of the URL.
-     */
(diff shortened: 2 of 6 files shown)


common-custom-user-data-gradle-plugin/src/main/java/com/gradle/Utils.java
```diff
@@ -105,6 +106,19 @@ static String urlEncode(String str) {
         }
     }
 
+    /**
+     * Redact any `user:password` that is part of the URL.
+     */
+    static String redactUserInfo(String url) {
```
etiennestuder, line 112: I would remove the javadoc. IMO, the method name says it all. (for Gradle and Maven)

common-custom-user-data-gradle-plugin/src/main/java/com/gradle/Utils.java
```diff
@@ -105,6 +106,19 @@ static String urlEncode(String str) {
...
+        try {
+            URI uri = new URI(url);
+            String userInfoPart = uri.getUserInfo() + '@';
+            return url.replace(userInfoPart, "******@");
```
etiennestuder, line 116: This does not seem to be null-safe (from the getUserInfo Javadoc). (for Gradle and Maven)

common-custom-user-data-gradle-plugin/CHANGELOG.md
```diff
@@ -4,6 +4,7 @@ All notable changes to this project will be documented in this file.
 ## [Unreleased]
 - Add custom value and search link for GitHub Actions run ID.
 - Fix generation of search links for custom values (#187).
```
etiennestuder, line 6: I think we are inconsistent in our linking to issues from the release notes items. Maybe we should jus stop doing that.  (for Gradle and Maven)

etiennestuder (changes requested):

common-custom-user-data-gradle-plugin/src/main/java/com/gradle/Utils.java
bigdaz: Wow, this was [totally handled in the original PR](https://github.com/gradle/gradle-enterprise-build-config-samples/pull/160/files#diff-52523462d2092ee6f6abe1aa4ada47ac1fd2195f039cd20574562c2688c51b5fR109-R111) and I missed it when copy-pasting into my solution. 🤦 .

common-custom-user-data-gradle-plugin/src/main/java/com/gradle/Utils.java
bigdaz: Done

common-custom-user-data-gradle-plugin/src/main/java/com/gradle/Utils.java
bigdaz: Fixed

common-custom-user-data-gradle-plugin/CHANGELOG.md
bigdaz: Reference to issue removed.

etiennestuder (approved):

PR #197, opened 2021-11-08
Use 'buildScan.buildFinished' for deferred actions

bigdaz:
To ensure that the correct build scan server is used when generating
custom value search links, we need to wait until no further changes will
be made to the server configuration.

Instead of `settingsEvaluated` and `afterEvalutate`, this change makes use
of `BuildScanExtension.buildFinished` to defer this work. This ensures that even
if the server URL is set during build execution the correct value will be used.

I considered keeping the mechanism to collect all of the different link-create actions into a single composite action to run with `buildFinished`. However, the mechanism for queuing multiple `buildFinished` actions is lightweight, so this seemed unnecessary.

### DIFF
--- a/common-custom-user-data-gradle-plugin/src/main/java/com/gradle/CommonCustomUserDataGradlePlugin.java
+++ b/common-custom-user-data-gradle-plugin/src/main/java/com/gradle/CommonCustomUserDataGradlePlugin.java
@@ -56,9 +56,6 @@ public class CommonCustomUserDataGradlePlugin implements Plugin<Object> {
                 SystemPropertyOverrides overrides = new SystemPropertyOverrides(providers);
                 overrides.configureGradleEnterprise(gradleEnterprise);
                 overrides.configureBuildCache(buildCache);
-
-                // do the enhancements that can only be done once all configuration has been fully applied
-                buildScanEnhancements.runPostEvaluateActions();
             });
         });
     }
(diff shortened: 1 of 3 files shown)


facewindu:
FWIW this is better and probably sufficient in 99% of cases, but this is not perfect.
We have _no_ callbacks that you can call once the configuraiton is locked (see https://github.com/gradle/dotcom/blob/d69589a29ef3d51419e50f787d0748a41268511a/build-agent-gradle/src/main/java/com/gradle/scan/plugin/internal/service/EndOfBuildListener.java#L33-L38)
Even with this change, you can do sth like
```
buildScan.buildFinished { 
  // capture the urls you want based on the GE server url
}
buildScan.buildFinished { 
  // change the GE server url
}
```
which will invalidate your captured URLs.
As stated in https://github.com/gradle/gradle-enterprise-build-config-samples/pull/192#issuecomment-962905469, we need a callback after locking for this kind of things

bigdaz:
@facewindu Yep thanks for the explanation. This PR is designed to make the CCUDs work as well as possible within the current build-agent limitations. 

@etiennestuder Would you like to review this as well?

etiennestuder (commented):
LGTM. Did you make the similar change for Maven already in a different PR?

etiennestuder:
| As stated in #192 (comment), we need a callback after locking for this kind of things

We can live with this for now.

PR #198, opened 2021-11-08
Build CCUD Gradle plugin with Java 8

bigdaz:
Use the 'java.toolchain' feature of Gradle to always build the plugin
with a Java 1.8 JDK.

This allows us to run the build with newer JDK versions, and will
make it easier to catch inadvertant use of newer JDK features/APIs.

We are already [doing the equivalent for the CCUD Maven extension](https://github.com/gradle/gradle-enterprise-build-config-samples/blob/master/common-custom-user-data-maven-extension/pom.xml#L90-L91).


etiennestuder (approved):
LGTM

PR #206, opened 2021-11-19
Fix for using other build cache types

runningcode:
When using a different build cache type the CCUD plugin would replace it with an HttpBuildCache. This checks the type to avoid replacement.

### DIFF
--- a/common-custom-user-data-gradle-plugin/src/main/java/com/gradle/SystemPropertyOverrides.java
+++ b/common-custom-user-data-gradle-plugin/src/main/java/com/gradle/SystemPropertyOverrides.java
@@ -53,7 +53,8 @@ final class SystemPropertyOverrides {
         });
 
         // null check required to avoid creating a remote build cache instance when none was already present in the build
-        if (buildCache.getRemote() != null) {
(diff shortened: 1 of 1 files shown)


common-custom-user-data-gradle-plugin/src/main/java/com/gradle/SystemPropertyOverrides.java
```diff
@@ -53,7 +53,8 @@ void configureBuildCache(BuildCacheConfiguration buildCache) {
         });
 
         // null check required to avoid creating a remote build cache instance when none was already present in the build
-        if (buildCache.getRemote() != null) {
+        // We need to ensure that this is the HttpBuildCache and not another build cache type like AWS S3.
+        if (buildCache.getRemote() != null && buildCache.getRemote() instanceof HttpBuildCache) {
```
anuraaga, line 57: Can remove null check which will be handled by instanceof

runningcode: done, thanks.

etiennestuder (approved):
LGTM

PR #211, opened 2021-11-25
Add GoCD support to the CCUD Gradle plugin and CCUD Maven extension

alextu:
Adds:
- CI tag
- custom links for getting build scans with same "pipeline", "stage", "job"

Tested manually locally on a GoCD instance with both Gradle and Maven builds.

### DIFF
--- a/common-custom-user-data-gradle-plugin/src/main/java/com/gradle/CustomBuildScanEnhancements.java
+++ b/common-custom-user-data-gradle-plugin/src/main/java/com/gradle/CustomBuildScanEnhancements.java
@@ -12,6 +12,7 @@ import java.util.Properties;
 import java.util.function.Supplier;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
+import java.util.stream.Stream;
 
 import static com.gradle.Utils.appendIfMissing;
 import static com.gradle.Utils.execAndCheckSuccess;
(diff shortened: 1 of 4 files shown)


common-custom-user-data-gradle-plugin/src/main/java/com/gradle/CustomBuildScanEnhancements.java
```diff
@@ -188,10 +188,21 @@ private void captureCiMetadata() {
             envVariable("BITRISE_BUILD_NUMBER").ifPresent(value ->
                 buildScan.value("CI build number", value));
         }
+
+        if (isGoCD()) {
+            envVariable("GO_SERVER_URL").ifPresent(url ->
+                buildScan.link("GoCD", url));
```
alextu, line 194: More useful would be the direct URL to the build but it's not provided as a direct env var, it could be built from other available env vars but those generated URLs could change in a future GoCD version...

bigdaz: We already construct URLs for [TeamCity](https://github.com/gradle/gradle-enterprise-build-config-samples/blob/56e4ab082ffb8887323ac7d791e3aa5134da229a/common-custom-user-data-gradle-plugin/src/main/java/com/gradle/CustomBuildScanEnhancements.java#L115) and [GitHub Actions](https://github.com/gradle/gradle-enterprise-build-config-samples/blob/56e4ab082ffb8887323ac7d791e3aa5134da229a/common-custom-user-data-gradle-plugin/src/main/java/com/gradle/CustomBuildScanEnhancements.java#L156). Would the logic for GoCD be a lot more complex, or the end result possibly more susceptible to change? 

I'd try to be consistent with other CI vendors unless there's a compelling reason not to be.

common-custom-user-data-gradle-plugin/src/main/java/com/gradle/CustomBuildScanEnhancements.java
```diff
@@ -188,10 +188,21 @@ private void captureCiMetadata() {
...
+            envVariable("GO_PIPELINE_NAME").ifPresent(value ->
+                addCustomValueAndSearchLink("CI pipeline name", value));
```
bigdaz, line 196: I think `addCustomValueAndSearchLink("CI pipeline", value))` would be better here. 
This will result in:
- Custom value `CI pipline`
- Link `CI pipeline build scans`

(It's also possible to specify the link label separately from the custom value name if required).

bigdaz (changes requested):
1. Would opt for constructed link to the actual build on GoCD, unless this is particularly problematic.
2. Should mention in CHANGELOG
3. Minor change requested to link/value name

common-custom-user-data-gradle-plugin/src/main/java/com/gradle/CustomBuildScanEnhancements.java
alextu: oh thanks I haven't noticed that, so I will do the same as GoCD URL seems simple enough:
`http://localhost:8153/go/tab/build/detail/<pipeline_name>/<pipeline_number>/<stage_name>/<stage_number>/<job_name>`

I don't know if those URLs will change anytime soon, my concern is that if they change we won't know be notified (until a user tell us :)), and it will be probably hard to fix for all CI versions with various URLs format.

bigdaz (approved):
LGTM

runningcode:
👍  Nice!

PR #220, opened 2021-12-16
Validate Maven build caching samples

jprinet:
Here is a GH workflow to validate our Maven build caching samples.

The current version is good enough to catch unknown fields in a Mojo.
An unhandled field is only logged as a warning in the console when a GE backend is effectively plugged. This is not supported at the moment.

Let me know @bigdaz if this is what you were thinking about.

### DIFF
--- a/.github/workflows/build-caching-maven-samples.yml
+++ b/.github/workflows/build-caching-maven-samples.yml
@@ -6,9 +6,6 @@ jobs:
 
     verify-build-caching-maven-samples:
         runs-on: ubuntu-latest
-        env:
-            LOG_MSG_EXTENSION_LOADED: "Build operation 'Gradle Enterprise mojo execution' completed"
-            LOG_MSG_EXECUTION_FAILED: "Error executing a GradleEnterpriseListener callback"
 
         steps:
             - name: Checkout
--- a/.github/workflows/build-caching-maven-samples.yml
+++ b/.github/workflows/build-caching-maven-samples.yml
@@ -23,4 +23,4 @@ jobs:
               run: npm install -g yarn
             - name: Run Maven verify
               working-directory: ./build-caching-maven-samples
-              run: ./mvnw clean verify -Dorg.slf4j.simpleLogger.log.gradle.goal.cache=debug -Dgradle.scan.disabled=true
+              run: ./mvnw clean verify -Dgradle.scan.disabled=true
--- a/.github/workflows/build-caching-maven-samples.yml
+++ b/.github/workflows/build-caching-maven-samples.yml
@@ -1,0 +1,29 @@
+name: Build caching Maven sample validation
+
+on: [push]
+
+jobs:
+
+    verify-build-caching-maven-samples:
+        runs-on: ubuntu-latest
+        env:
+            LOG_MSG_EXTENSION_LOADED: "Build operation 'Gradle Enterprise mojo execution' completed"
+            LOG_MSG_EXECUTION_FAILED: "Error executing a GradleEnterpriseListener callback"
+
+        steps:
+            - name: Checkout
+              uses: actions/checkout@v2
+            - name: Set up JDK 11
+              uses: actions/setup-java@v2
+              with:
+                  java-version: '11'
+                  distribution: 'adopt'
+            - name: Setup Nodejs and npm
+              uses: actions/setup-node@v2
+              with:
+                  node-version: "14"
+            - name: Setup yarn
+              run: npm install -g yarn
+            - name: Run Maven verify
+              working-directory: ./build-caching-maven-samples
+              run: ./mvnw clean verify -Dorg.slf4j.simpleLogger.log.gradle.goal.cache=debug -Dgradle.scan.disabled=true


.github/workflows/build-caching-maven-samples.yml
```diff
@@ -0,0 +1,29 @@
...
+
+on: [push]
+
+jobs:
+
+    verify-build-caching-maven-samples:
+        runs-on: ubuntu-latest
+        env:
+            LOG_MSG_EXTENSION_LOADED: "Build operation 'Gradle Enterprise mojo execution' completed"
+            LOG_MSG_EXECUTION_FAILED: "Error executing a GradleEnterpriseListener callback"
```
bigdaz, line 11: Where are these variables consumed? What impact do they have?

.github/workflows/build-caching-maven-samples.yml
```diff
@@ -0,0 +1,29 @@
...
+                  distribution: 'adopt'
+            - name: Setup Nodejs and npm
+              uses: actions/setup-node@v2
+              with:
+                  node-version: "14"
+            - name: Setup yarn
+              run: npm install -g yarn
+            - name: Run Maven verify
+              working-directory: ./build-caching-maven-samples
+              run: ./mvnw clean verify -Dorg.slf4j.simpleLogger.log.gradle.goal.cache=debug -Dgradle.scan.disabled=true
```
bigdaz, line 29: How does this verify that build caching is correctly configured?

bigdaz (changes requested):
@jprinet I don't really understand what this workflow will check for, and how it works?
My preference would be to run the build connected to a real GE instance (so that local caching can be enabled).
The workflow would then effectively run Experiment 1 and verify the expected number of goals are FROM-CACHE.

jprinet:
> @jprinet I don't really understand what this workflow will check for, and how it works? My preference would be to run the build connected to a real GE instance (so that local caching can be enabled). The workflow would then effectively run Experiment 1 and verify the expected number of goals are FROM-CACHE.

This actual workflow raises a failure if an unknown property is set in the plugin customization (in addition to any global failure due to syntax for instance). This can arise with a new version.
I do agree that a great improvement would be to connect to GE to get unhandled property and check caching.

.github/workflows/build-caching-maven-samples.yml
jprinet: just a wrong copy/paste, deleting that

.github/workflows/build-caching-maven-samples.yml
jprinet: it does not, build cache is disabled if no GE instance is connected

bigdaz:
> This actual workflow raises a failure if an unknown property is set in the plugin customization (in addition to any global failure due to syntax for instance). This can arise with a new version.

OK got it. In my experience the most common issue is that the new Mojo version will have additional parameters, causing caching to stop working without any build failure. I'm OK with merging this PR as a first step, but I don't think it will catch the majority of issues that come with upgrading dependency versions.

jprinet:
my impression is that added value is not much while we don't have a way to get the caching feature enabled.
I would keep that on hold before having a more complete solution.
wdyt @etiennestuder ?

bigdaz:
@jprinet Do you know if e.grdev.net has Maven functionality enabled? If so, it would be nice to see if it could be used as a backing GE server for this repo. If it's possible, then we'd simply need to get approval to use e.grdev.net for this public repo: I'm not sure how important it is not to advertise that URL.

runningcode:
To summarize:
My understanding is that this will test for a property being configured in the Gradle Enterprise configuration which does not actually exist in the Mojo.
This will not test for a new property being added in - for example - an update unless we connect it to a GE instance that has Maven caching enabled.

While the added functionality is not much, it does allow us to expand this in the future and connect it to a GE instance for proper testing.

After discussing with @etiennestuder we came to this conclusion: let's remove the functionality to connect the build to a GE instance with secrets and only merge the parts that check for an unknown property. Once that is merged we can circle back on adding the rest of the functionality.

.github/workflows/build-caching-maven-samples.yml
runningcode: can we remove the `-Dorg.slf4j.simpleLogger.log.gradle.goal.cache=debug` line?

.github/workflows/build-caching-maven-samples.yml
```diff
@@ -0,0 +1,26 @@
...
+            - name: Set up JDK 11
+              uses: actions/setup-java@v2
+              with:
+                  java-version: '11'
+                  distribution: 'adopt'
+            - name: Setup Nodejs and npm
+              uses: actions/setup-node@v2
+              with:
+                  node-version: "14"
+            - name: Setup yarn
```
runningcode, line 22: why do we need yarn installed?

jprinet:
Reverted to the version without requirement for a GE backend, please confirm that this is good to go @runningcode

.github/workflows/build-caching-maven-samples.yml
jprinet: `exec-yarn-project` requires it

runningcode: ah, of course!

runningcode (approved):
Looks great! Feel free to merge.

PR #230, opened 2021-12-22
Make Github actions more consistent between Maven and Gradle

jprinet:
Objective is to have Gradle and Maven GH actions more consistent.
There are still some differences as configuration and verification steps are slightly different for Maven and Gradle.


etiennestuder:
@jprinet Why is this isse closed before being reviewed?

jprinet:
Sorry for the noise @etiennestuder 
Last minute review made me think about an improvement, no need to review the current version.

PR #231, opened 2021-12-22
Make Github actions more consistent between Maven and Gradle

jprinet:
Objective is to have Gradle and Maven GH actions more consistent.
There are still some differences as configuration and verification steps are slightly different for Maven and Gradle.

### DIFF
--- a/.github/workflows/gradle-smoke-test.yml
+++ b/.github/workflows/gradle-smoke-test.yml
@@ -10,17 +10,17 @@ jobs:
         strategy:
             matrix:
                 include:
-                    - project-dir: "common-custom-user-data-gradle-plugin"
+                    - build-dir: "common-gradle-enterprise-gradle-configuration"
                       sample-file: "../build-data-capturing-gradle-samples/capture-dependency-resolution/gradle-dependency-resolution.gradle"
-                    - project-dir: "common-custom-user-data-gradle-plugin"
+                    - build-dir: "common-gradle-enterprise-gradle-configuration"
                       sample-file: "../build-data-capturing-gradle-samples/capture-git-diffs/gradle-git-diffs.gradle"
-                    - project-dir: "common-custom-user-data-gradle-plugin"
+                    - build-dir: "common-gradle-enterprise-gradle-configuration"
                       sample-file: "../build-data-capturing-gradle-samples/capture-os-processes/gradle-os-processes.gradle"
-                    - project-dir: "common-custom-user-data-gradle-plugin"
+                    - build-dir: "common-gradle-enterprise-gradle-configuration"
                       sample-file: "../build-data-capturing-gradle-samples/capture-quality-check-issues/gradle-quality-check-issues.gradle"
-                    - project-dir: "common-custom-user-data-gradle-plugin"
+                    - build-dir: "common-gradle-enterprise-gradle-configuration"
                       sample-file: "../build-data-capturing-gradle-samples/capture-slow-workunit-executions/gradle-slow-task-executions.gradle"
-                    - project-dir: "common-custom-user-data-gradle-plugin"
+                    - build-dir: "common-gradle-enterprise-gradle-configuration"
                       sample-file: "../build-data-capturing-gradle-samples/capture-test-execution-system-properties/gradle-test-execution-system-properties.gradle"
 
         steps:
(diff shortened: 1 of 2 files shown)


.github/workflows/gradle-smoke-test.yml
```diff
@@ -40,19 +31,34 @@ jobs:
...
-              # - Add a settings.gradle
+            - name: Publish current snapshot of CCUD Gradle plugin to Maven local repository
+              uses: gradle/gradle-build-action@v2
+              with:
+                  arguments: publishToMavenLocal -x signArchives
+                  build-root-directory: common-custom-user-data-gradle-plugin
+            - name: Replace version in build configuration
+              run: |
+                  # fetching current snapshot version
+                  CURRENT_SNAPSHOT_VERSION=$(awk '/version =/ {print $3;exit;}' common-custom-user-data-gradle-plugin/build.gradle)
```
runningcode, line 42: instead of using a snapshot version published to a local maven repository, can we use the composite build command line flag `--include-build`? https://docs.gradle.org/current/userguide/composite_builds.html#command_line_composite

this could remove the need for the script below that adds the `mavenLocal` to the `settings.gradle` file.

.github/workflows/gradle-smoke-test.yml
```diff
@@ -3,34 +3,25 @@ name: Gradle smoke test
...
         strategy:
             matrix:
                 include:
-                    - path-to-sample-build-file: "build-data-capturing-gradle-samples/capture-dependency-resolution/gradle-dependency-resolution.gradle"
-                    - path-to-sample-build-file: "build-data-capturing-gradle-samples/capture-git-diffs/gradle-git-diffs.gradle"
-                    - path-to-sample-build-file: "build-data-capturing-gradle-samples/capture-os-processes/gradle-os-processes.gradle"
-                    - path-to-sample-build-file: "build-data-capturing-gradle-samples/capture-quality-check-issues/gradle-quality-check-issues.gradle"
-                    - path-to-sample-build-file: "build-data-capturing-gradle-samples/capture-slow-workunit-executions/gradle-slow-task-executions.gradle"
-                    - path-to-sample-build-file: "build-data-capturing-gradle-samples/capture-test-execution-system-properties/gradle-test-execution-system-properties.gradle"
+                    - build-dir: "common-gradle-enterprise-gradle-configuration"
```
runningcode, line 13: Where do we test the `common-custom-user-data-gradle-plugin`?

jprinet: We are building it in the "Publish current snapshot of the CCUD Gradle plugin to local Maven repository" step and then using it on the samples

.github/workflows/gradle-smoke-test.yml
jprinet: I thought about this approach, but our instructions in the samples are to use the `apply from` syntax.
The other argument is to have Maven and Gradle steps alike.

This being said I'm ok with the included build approach, let me know your preference.

.github/workflows/gradle-smoke-test.yml
runningcode: Ah that makes sense since there aren't any tests in there.

.github/workflows/gradle-smoke-test.yml
runningcode: I didn't mean for the samples, I meant for this build specifically, instead of having a multi-part publish to maven local and then consume from maven local, we could simplify the build by having a single step included build. Can you try this out and see if it simplifies this build yaml file?

bigdaz: I agree with @runningcode that an included build would be more appropriate here. You should be able to eliminate `publishToMavenLocal`, locating and replacing the plugin version number, and adding `mavenLocal` as a plugin repository by simply running the build with: `./gradlew tasks --include-build ../common-custom-user-data-gradle-plugin`.

One downside is that we can't do something similar with the CCUDME, so the GitHub Actions won't be consistent between Maven and Gradle. But I think it's simpler and more idiomatic for Gradle.

jprinet: all good then, I'm preparing this!

runningcode: @bigdaz I think @jprinet is running in to a limitation where the `--include-build` approach does not make substitutions for settings plugins. :\

jprinet: if you can't get the substitution through an included build, I'm gently asking your review again on the current version ;)

runningcode (approved):
This looks good to me. Thanks for making this more consistent @jprinet !

jprinet:
> This looks good to me. Thanks for making this more consistent @jprinet !

Can you please confirm @bigdaz that this is good to go?

runningcode:
@jprinet Let's merge this now and Daz can suggest any improvements in the future.

PR #237, opened 2021-12-31
Tag build when various skip tests property are set to true or empty

alextu:
Set specific tag(s) regarding only system properties `skipTests`, `maven.test.skip`, `skipITs`. Note that tests can be skipped with surefire/failsafe specific configuration, this is not taken into account as it can be specific per project. Here's a recap of various cases:

```        
        label                                                         | system prop               | expectedTag
        --------------------------------------------------------------------------------------------------------
        "skip all tests (no value)"                                   | "-DskipTests"             | "skipTests"
        "skip all tests"                                              | "-DskipTests=true"        | "skipTests"
        "do not skip all tests"                                       | "-DskipTests=false"       | _
        "do not skip all tests (unknown value)"                       | "-DskipTests=foo"         | _
        "skip all integration tests (no value)"                       | "-DskipITs"               | "skipITs"
        "skip all integration tests"                                  | "-DskipITs=true"          | "skipITs"
        "do not skip all integration tests"                           | "-DskipITs=false"         | _
        "do not skip all integration tests (unknown value)"           | "-DskipITs=foo"           | _
        "skip compiling and running all tests (no value)"             | "-Dmaven.test.skip"       | "skipTests"
        "skip compiling and running all tests"                        | "-Dmaven.test.skip=true"  | "skipTests"
        "do not skip compiling and running all tests"                 | "-Dmaven.test.skip=false" | _
        "do not skip compiling and running all tests (unknown value)" | "-Dmaven.test.skip=foo"   | _
```

https://github.com/gradle/ge/issues/11242

### DIFF
--- a/common-custom-user-data-maven-extension/src/main/java/com/gradle/CustomBuildScanEnhancements.java
+++ b/common-custom-user-data-maven-extension/src/main/java/com/gradle/CustomBuildScanEnhancements.java
@@ -42,6 +42,21 @@ final class CustomBuildScanEnhancements {
         captureCiOrLocal();
         captureCiMetadata();
         captureGitMetadata();
+        captureSkipTests();
+    }
+
+    private void captureSkipTests() {
+        customValueWhenTrueOrEmpty("skipTests");
+        customValueWhenTrueOrEmpty("maven.test.skip");
+        customValueWhenTrueOrEmpty("skipITs");
+    }
+
+    private void customValueWhenTrueOrEmpty(String property) {
+        projectProperty(property).ifPresent(value -> {
+            if (Boolean.TRUE.toString().equals(value) || value.isEmpty()) {
+                buildScan.value("maven.switches", property);
+            }
+        });
     }
 
     private void captureOs() {


etiennestuder:
I would
- add this code as a snippet to build-data-capturing-gradle-samples as proposed by @runningcode 
- I would differentiate between `skip all tests` and `skip compiling and running all tests` in the tag names (or add two tags?)

> this is not taken into account as it can be specific per project

I do not understand why you mean.

alextu:
> Note that tests can be skipped with surefire/failsafe specific configuration, this is not taken into account as it can be specific per project. 

Adding precisions to this:

In maven, you can skip tests from a specific project by configuring the surefire goal on this project pom file: `<configuration><skipTests>true</skipTests></configuration>`. In this case it won't be detected by the CCUD mvn extension or sample code and no tag will be added. Moreover this configuration flag takes precedence over the `skipTests` system property, so:
- a build scan not tagged with `skipTests` does not mean all tests have run, some may have been skipped
- a build scan tagged with `skipTests` does not mean all tests have been skipped, some may have run 

Bottom line, the tag just means "this build has been run with `-DskipTests`" (it's a similar behavior with `skipITs`)

bigdaz:
> Thanks for doing this. I'm not sure this is something that we want to add to all builds. I think this makes more sense as a build-data-capturing-gradle-samples.

A counter-opinion: I found in trials, when some builds were run with `-DskipTests`, that this was quite confusing and hard to diagnose. I think we should always have a tag to make this clear, until such a differentiator is added directly to Gradle Enterprise.

The equivalent in Gradle would be if you couldn't filter between builds running `gradle assemble` or `gradle build`: it seems quite common to use `mvn clean verify -DskipTests` and `mvn clean verify` as the equivalent commands in Maven.

I would rate this tag as more useful than the tag for git branch name or dirty status. Just my 2c: I won't fight for it.

runningcode:
I see your point Daz. I think there are a lot of cases where command line argument are extremely useful and can completely change the build outcome in hard to diagnose ways. We don't capture all command line arguments right now in GE for either Gradle or Maven builds.
The thing is, we could start adding these command line argument to CCUD, but at what point do we stop? 

@etiennestuder Perhaps this is better addressed by adding this to the switches section in the GE Maven extension? We capture similar values like `--dry-run` for Gradle builds.

alextu:
After our discussion during the GE sync, here's the conclusion:
- keep the code in the CCUD maven extension
- set custom values instead of tags with the following suggested name/value:
```
 label                                                                | system prop               | custom value
        --------------------------------------------------------------------------------------------------------
        "skip all tests (no value)"                                   | "-DskipTests"             | "skipTests = true"
        "skip all tests"                                              | "-DskipTests=true"        | "skipTests = true"
        "do not skip all tests"                                       | "-DskipTests=false"       | "skipTests = false"
        "do not skip all tests (unknown value)"                       | "-DskipTests=foo"         | _
        "skip all integration tests (no value)"                       | "-DskipITs"               | "skipITs = true"
        "skip all integration tests"                                  | "-DskipITs=true"          | "skipITs = true"
        "do not skip all integration tests"                           | "-DskipITs=false"         | "skipITs = false"
        "do not skip all integration tests (unknown value)"           | "-DskipITs=foo"           | _
        "skip compiling and running all tests (no value)"             | "-Dmaven.test.skip"       | "maven.test.skip=true"
        "skip compiling and running all tests"                        | "-Dmaven.test.skip=true"  | "maven.test.skip=true"
        "do not skip compiling and running all tests"                 | "-Dmaven.test.skip=false" | "maven.test.skip=false"
        "do not skip compiling and running all tests (unknown value)" | "-Dmaven.test.skip=foo"   | _
```
The idea is to keep the same naming as the original maven flags, but still translate a bit the value(s) so they can be searchable without having to remember all possible values that could have been entered. @etiennestuder does it sound about right ?

etiennestuder:
One thing we did not conclude is whether to also capture custom values in case the flag is OFF. In my opinion, we should only capture the value, if it resolves to true, to avoid too much noise in the custom values section.

etiennestuder:
Another thing to consider is how we capture these values. An alternative would be:

maven.switches=skipTests
maven.switches=skipITs
maven.switches=maven.test.skip

(note that you can have multiple values with the same name)

The reason for bringing up this alternative is that it would keep them together in the UI, and it would scale to other types of switches we might capture in the future.

alextu:
Ok that makes more sense, so summarizing here:
- setting `maven.switches=skipTests` if `-DskipTests` or `-DskipTests=true`
- same goes for the other 2 flags `skipITs` and `maven.test.skip`
I will make the corresponding changes today

alextu:
Actually, specifying 2 custom values with the same name, for example `./mvnw clean verify -Dscan.value.maven.switches=value1 -Dscan.value.maven.switches=value2` results in only setting one custom value `maven.switches=value2` (tried also with the API `buildScan.value`). @etiennestuder were you referring to another way of setting multiple times the same custom value ? Maybe should we set a custom value with the various comma seperated values like: `maven.switches=value1,value2` ?

alextu:
My bad, it actually works when using the API, not with the system property which makes sense since Java system properties are stored in a map

etiennestuder:
The alternative would be to do:

`maven.switches.skipTests=true`
`maven.switches.skipITs=true`
etc.

I think I like this one best.

alextu:
@etiennestuder I've updated the PR with the first suggestion: `maven.switches=<skip_test_flag>`, you can check the corresponding tests here: https://github.com/gradle/ge/compare/atual/mvn-ccud-skiptests?expand=1#diff-38cd3883c0943eab12be990ff92230f0855459fd1515cf53d104a6cd7b8a546d (to be merged after merging this PR)

etiennestuder:
I'll take a look this morning.

PR #249, opened 2022-01-14
Add folders to dependabot configuration

jprinet:
fixing _build-caching-maven-samples_ and adding _rollout-maven-extension_ and _gradle-enterprise-conventions-gradle-plugin_ to be monitored by `dependabot`

Please let me know @bigdaz  if there were good reasons not to do so 😅

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -7,7 +7,7 @@ registries:
     password: dummy # Required by dependabot
 updates:
   - package-ecosystem: "maven"
-    directory: "maven-build-caching-samples"
+    directory: "build-caching-maven-samples"
     schedule:
       interval: "daily"
   - package-ecosystem: "maven"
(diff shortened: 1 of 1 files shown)


etiennestuder:
Nice, @jprinet . I will take a look soon.

bigdaz:
No objection from me. Just some comments:
- For `build-caching-maven-samples` this looks good. We failed to change the dependabot config when this was renamed recently, so 👍 
- For `rollout-maven-extension`, this will only have an impact if `.mvn/extensions.xml` is supported by dependabot. I guess there's no harm even if it in not yet supported.
- I'm not really familiar with `gradle-enterprise-conventions-gradle-plugin`, but it looks like we'll need to watch to avoid unintentionally upgrading [dependencies like this](https://github.com/gradle/gradle-enterprise-build-config-samples/blob/master/gradle-enterprise-conventions-gradle-plugin/plugins/gradle-2-through-4/build.gradle#L15).

etiennestuder:
> * For `rollout-maven-extension`, this will only have an impact if `.mvn/extensions.xml` is supported by dependabot. I guess there's no harm even if it in not yet supported.

It is supported.



> * I'm not really familiar with `gradle-enterprise-conventions-gradle-plugin`, but it looks like we'll need to watch to avoid unintentionally upgrading [dependencies like this](https://github.com/gradle/gradle-enterprise-build-config-samples/blob/master/gradle-enterprise-conventions-gradle-plugin/plugins/gradle-2-through-4/build.gradle#L15).

Good point. I would not include this one. It will lile move somewhere else soon anyway.

PR #293, opened 2022-02-28
Publish scans to ge.solutions-team.gradle.com

runningcode:


build-caching-maven-samples/.mvn/gradle-enterprise.xml
```diff
@@ -1,17 +1,21 @@
-<gradleEnterprise>
+<?xml version="1.0" encoding="UTF-8" standalone="yes" ?>
+
+<gradleEnterprise
+        xmlns="https://www.gradle.com/gradle-enterprise-maven" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xsi:schemaLocation="https://www.gradle.com/gradle-enterprise-maven https://www.gradle.com/schema/gradle-enterprise-maven.xsd">
     <server>
-        <url>https://localhost</url>
-        <allowUntrusted>true</allowUntrusted>
+        <url>https://ge.solutions-team.gradle.com</url>
```
runningcode, line 7: This has the effect of potentially addressing https://github.com/gradle/ge/issues/13226 although I did not verify that the builds will fail with incorrect samples.

jprinet: I think we should take advantage of this, running a second build and making sure the number of cache hits remain constant.

runningcode:
We will keep these as samples and not merge this change to point to our solutions instance.

PR #305, opened 2022-03-17
Add wrapper upgrade

alextu:


.github/workflows/wrapper-upgrade-execution.yml
```diff
@@ -0,0 +1,41 @@
...
+          passphrase: ${{ secrets.GH_BOT_PGP_PASSPHRASE }}
+          git_user_signingkey: true
+          git_commit_gpgsign: true
+          git_config_global: true
+      - name: Set up JDK 11
+        uses: actions/setup-java@v2
+        with:
+          java-version: '11'
+          distribution: 'adopt'
+      - name: Clean with Gradle
```
etiennestuder, line 30: IMO, we can remove the clean step.

.github/workflows/wrapper-upgrade-execution.yml
```diff
@@ -0,0 +1,41 @@
...
+        uses: gradle/gradle-build-action@v2
+        with:
+          build-root-directory: wrapper-upgrade
+          arguments: 'clean'
+      - name: Upgrade Gradle Wrappers
+        uses: gradle/gradle-build-action@v2
+        with:
+          build-root-directory: wrapper-upgrade
```
etiennestuder, line 38: That directory does not seem right.

.github/workflows/wrapper-upgrade-execution.yml
```diff
@@ -0,0 +1,41 @@
...
+          arguments: 'upgradeGradleWrapperAll --continue'
```
etiennestuder, line 39: There is also a Maven wrapper in this repo.

build.gradle
```diff
@@ -0,0 +1,25 @@
...
+        'gradle-enterprise-conventions-gradle-plugin-gradle2' {
+            repo = 'gradle/gradle-enterprise-build-config-samples'
+            dir = 'gradle-enterprise-conventions-gradle-plugin/plugins/gradle-2-through-4'
+        }
+        'gradle-enterprise-conventions-gradle-plugin-gradle5' {
+            repo = 'gradle/gradle-enterprise-build-config-samples'
+            dir = 'gradle-enterprise-conventions-gradle-plugin/plugins/gradle-5-or-newer'
+        }
+    }
+}
```
etiennestuder, line 25: The Maven repo is missing.

etiennestuder (changes requested):

.github/workflows/wrapper-upgrade-execution.yml
alextu: Yes, but still needs to execute a clean since so our wrapper upgrade plugin will be able to clone (if not it fails because the directory already exists in `build/git-clones`

build.gradle
alextu: Good catch

.github/workflows/wrapper-upgrade-execution.yml
etiennestuder: I don't understand.

alextu: I've removed the `clean`, it should not be a problem if GH action state restore does not interfere

etiennestuder (approved):

PR #330, opened 2022-04-01
Bump up GE Maven extension to 1.13.1

jprinet:


rollout-maven-extension/.mvn/extensions.xml
```diff
@@ -7,7 +7,7 @@
     <extension>
         <groupId>com.gradle</groupId>
         <artifactId>gradle-enterprise-maven-extension</artifactId>
-        <version>1.13</version>
+        <version>1.13.1</version>
```
etiennestuder, line 10: @jprinet This should be caught by dependabot. Were you just faster, or are we missing this entry in the dependaboy yaml? (I believe the former is the case)

jprinet: The entry is listed in the `dependabot` configuration @etiennestuder , but other PR were already raised at that time
https://github.com/gradle/gradle-enterprise-build-config-samples/pulls?q=is%3Apr+is%3Aclosed+1.13.1

Let's monitor this on next upgrade.

PR #332, opened 2022-04-04
Add check for unknown Mojo params in verification workflow

alextu:
Will fail the `maven-build-caching-samples-verification.yml`  workflow if, for example, after a maven plugin update a new param is introduced and not mapped by our samples.

Note that the present check is rather naive, a more complete check would involve querying GE api but that's something @jprinet is working on with running validation scripts with a GH action. His work could be re-used there once done.

### DIFF
--- a/.github/workflows/maven-build-caching-samples-verification.yml
+++ b/.github/workflows/maven-build-caching-samples-verification.yml
@@ -6,6 +6,8 @@ jobs:
   verification:
     name: Verification
     runs-on: ubuntu-latest
+    env:
+      GRADLE_ENTERPRISE_ACCESS_KEY: ${{ secrets.GE_SOLUTIONS_ACCESS_TOKEN }}
     steps:
       - name: Checkout
         uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


jprinet:
The PR definitely addresses the check for unknown properties but my impression is that we could go a bit further running an actual E2E test, having a first build to seed the remote cache and a second to download from it.
This would have the benefit of checking for any unexpected cache miss at the price of an extra step.

.github/workflows/maven-build-caching-samples-verification.yml
```diff
@@ -21,5 +23,11 @@ jobs:
       - name: Set up Yarn
         run: npm install -g yarn
       - name: Build with Maven
+        id: mvn-build
         working-directory: ./build-caching-maven-samples
-        run: ./mvnw clean verify -Dgradle.scan.disabled=true
+        run: |
+          ./mvnw clean verify -Dgradle.enterprise.url=https://ge.solutions-team.gradle.com -Dorg.slf4j.simpleLogger.log.gradle.goal.cache=debug 2>&1 | tee -a /tmp/gradle-build.log
```
jprinet, line 29: I would have the GE URL as Secret

etiennestuder: @jprinet the URL of the Solutions Team's GE instance is public. No need to hide it.

runningcode (approved):
This is an excellent simple example that achieves what we want. We can always add extra functionality later. Thanks @alextu

alextu:
> The PR definitely addresses the check for unknown properties but my impression is that we could go a bit further running an actual E2E test, having a first build to seed the remote cache and a second to download from it. This would have the benefit of checking for any unexpected cache miss at the price of an extra step.

I think here using local cache only would be fine, it would require:
- cleaning the local cache
- running the build
- re-running the build
- checking with the rest api that all cacheable tasks have not executed

I can do it but it's probably very similar to what you're doing with the GH action running build validation scripts, so would probably be to re-use it once ready

PR #338, opened 2022-04-11
Add java toolchain to gradle convention plugin

alextu:


etiennestuder:
I would not use Java toolchain in this specific example. It serves more as copy/paste or inspirational than too specifically build something.

PR #341, opened 2022-04-12
Add "Set up Gradle" step to GHActions workflows

bigdaz:
Separating the setup of the Gradle build tool from actual build invocation
is more idiomatic to GH Actions in general, and provides a clearer separation
of responsibilities.


runningcode (approved):
👍

alextu (approved):

PR #443, opened 2022-08-24
Fix PMD Maven plugin caching sample

alextu:
Version `3.18.0` of the Maven PMD plugin had some changes causing our sample to fail the build:
- renamed property: `sourceEncoding` to `inputEncoding`
- new properties: `localRepository` and `remoteRepositories`


runningcode (approved):
Thanks for fixing this!

PR #460, opened 2022-09-19
Remove usage of GITHUB_ENV from workflow, instead prefer to use output parameters.

ryandens:


runningcode (approved):
Thanks @ryandens

jprinet (approved):
LGTM

PR #461, opened 2022-09-20
Capture thermal throttling samples

jprinet:


runningcode:
@ghale could you take a pass at this given you started some of the original thermal throttling scripts?

ghale (approved):
Looks good.  One question about the debug_all flag, but I don't think that should block this unless you want to change it.

runningcode (commented):
Is this sample configuration cache compatible?

build-data-capturing-gradle-samples/capture-thermal-throttling/gradle-thermal-throttling.gradle
```diff
@@ -0,0 +1,150 @@
...
+                  [level: "THROTTLING_HIGH", range: 0..40],
+                  [level: "THROTTLING_MEDIUM", range: 40..80],
+                  [level: "THROTTLING_LOW", range: 80..100]
+          ]
+
+  /**
+   * Flag to dump all captured values as custom values.
+   * Enable cautiously in case of very large sample collection, this can be adjusted with sampling interval.
+   */
+  def static final DEBUG_ALL_VALUES = false
```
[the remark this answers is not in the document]
jprinet: That could be interesting to check for distribution on a long running build. A single value might be not enough in such case to understand the impact.
I'm keen removing it.

jprinet:
> Is this sample configuration cache compatible?

No it's not, this is mentioned in the header

build-data-capturing-gradle-samples/capture-thermal-throttling/gradle-thermal-throttling.gradle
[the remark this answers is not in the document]
jprinet: I removed it 👍

jprinet:
> > Is this sample configuration cache compatible?
> 
> No it's not, this is mentioned in the header

It is now CC compatible @runningcode, using GE plugin's `buildFinished` hook

build-data-capturing-gradle-samples/capture-thermal-throttling/gradle-thermal-throttling.gradle
```diff
@@ -0,0 +1,139 @@
...
+
+  def static final COMMAND_ARGS = ["pmset", "-g", "therm"]
+  def static final COMMAND_OUTPUT_PARSING_PATTERN = /CPU_Speed_Limit\s+=\s+/
+
+  def static final scheduler = Executors.newScheduledThreadPool(1)
+  def static final samples = new ConcurrentLinkedQueue<Integer>()
+
+  static {
+    if (OperatingSystem.current().macOsX) {
+      scheduler.scheduleAtFixedRate(new ProcessRunner(COMMAND_ARGS, ThermalThrottlingService::processCommandOutput), 0, SAMPLING_INTERVAL_IN_SECONDS, TimeUnit.SECONDS)
```
runningcode, line 68: @ghale it feels to me that having the scheduler start inside a `static` initializer block might cause some situations where this is started but never closed. Is that possible or is there a better way to do this like inside the creation of this class?

build-data-capturing-gradle-samples/capture-thermal-throttling/gradle-thermal-throttling.gradle
```diff
@@ -0,0 +1,139 @@
...
+}
+
+// Register Thermal throttling service
+def throttlingServiceProvider = gradle.sharedServices.registerIfAbsent("thermalThrottling", ThermalThrottlingService.class) {}
+
+// Start Thermal throttling service
+throttlingServiceProvider.get()
+
+// Process results on build completion
+buildScan {
```
runningcode, line 36: let's make this more similar to the Maven groovy code with `buildScan.buildFinished` together on the same line

jprinet: ✅

build-data-capturing-gradle-samples/capture-thermal-throttling/gradle-thermal-throttling.gradle
ghale: I suppose someone could modify the script and create such a situation.  I think it was originally written with the scheduler started in the constructor.  @jprinet was there a particular reason this was moved inside of a static block?  If not, I'd say just move it back to the constructor.

jprinet: I wanted to make sure we would start to capture data as early as possible.
Just to be on the same page here, what's the concern? Encountering an error during object creation and having the scheduler already started?

jprinet: I moved the block to the constuctor which has the side benefit of allowing methods to be non static

build-data-capturing-gradle-samples/capture-thermal-throttling/gradle-thermal-throttling.gradle
```diff
@@ -0,0 +1,139 @@
...
+  def static final COMMAND_ARGS = ["pmset", "-g", "therm"]
+  def static final COMMAND_OUTPUT_PARSING_PATTERN = /CPU_Speed_Limit\s+=\s+/
+
+  def final scheduler
+  def final samples
+
+  ThermalThrottlingService() {
+    scheduler = Executors.newScheduledThreadPool(1)
+    samples = new ConcurrentLinkedQueue<Integer>()
+    if (OperatingSystem.current().macOsX) {
```
runningcode, line 67: should we check this at the very beginning before registering the build service? this way absolutely no build logic is added on unsupported platforms?

build-data-capturing-gradle-samples/capture-thermal-throttling/gradle-thermal-throttling.gradle
```diff
@@ -0,0 +1,139 @@
...
+  throttlingServiceProvider.get().processResults(buildScanApi)
+}
+
+// Thermal Throttling service implementation
+abstract class ThermalThrottlingService implements BuildService<BuildServiceParameters.None>, AutoCloseable {
+
+  /**
+   * Sampling interval can be adjusted according to total build time.
+   */
+  def static final SAMPLING_INTERVAL_IN_SECONDS = 5
```
runningcode, line 46: are the `final` here needed? our other scripts don't use `final` or `static`? I think it would be good to keep them in the same consistent simple style.

build-data-capturing-maven-samples/capture-thermal-throttling/maven-thermal-throttling.groovy
```diff
@@ -0,0 +1,134 @@
...
+  def final samples
+
+  ThermalThrottlingService() {
+    scheduler = Executors.newScheduledThreadPool(1)
+    samples = new ConcurrentLinkedQueue<Integer>()
+  }
+
+  void start() {
+    def osName = System.getProperty("os.name")
+    if(osName.contains("OS X") || osName.startsWith("Darwin")) {
```
runningcode, line 63:
```suggestion
    if (osName.contains("OS X") || osName.startsWith("Darwin")) {
```

runningcode: also same comment here, can we quickly abort the script if we aren't on osX as opposed to checking in here?

runningcode (commented):
getting very close! :)

build-data-capturing-gradle-samples/capture-thermal-throttling/gradle-thermal-throttling.gradle
jprinet: that's a good point, let me adjust this

build-data-capturing-gradle-samples/capture-thermal-throttling/gradle-thermal-throttling.gradle
jprinet: It's an habit to ensure [safe initialization](https://docs.oracle.com/javase/specs/jls/se8/html/jls-17.html#jls-17.5)
I think we can remove the keywords to get an alignment, the concurrency shouldn't be much of an issue for this managed object.

runningcode: Yes, I think we should remove these to simplify these scripts and keep the style consistent.

jprinet: > Yes, I think we should remove these to simplify these scripts and keep the style consistent.

Agreed, this is done in local.  
Unrelated, but I discovered that when CC is restored the service is not triggered, I'm investigating this before moving forward

jprinet: I eventually encapsulated the service start in a `buildScan.background` block to make sure it would be triggered at the beginning of the build in any situation (without CC, with CC when storing entry, with CC when restoring entry).

Having the start service instruction at the root level of the script doesn't work when restoring a CC entry, in this case the service is only instantiated when the service is required in the `buildScan.buildFinished` block.

runningcode (approved):
LGTM, thanks for following up on all the details.

jprinet:
> oh! good to check the configuration cache compatibility. if the final is a requirement for the configuration cache, then let's keep the final keyword.

You're commenting an outdated version of the code, there is no `final` nor `static` anymore

build-data-capturing-gradle-samples/capture-thermal-throttling/gradle-thermal-throttling.gradle
```diff
@@ -30,12 +30,17 @@ if (OperatingSystem.current().macOsX) {
   // Register Thermal throttling service
   def throttlingServiceProvider = gradle.sharedServices.registerIfAbsent("thermalThrottling", ThermalThrottlingService.class) {}
 
-  // Start Thermal throttling service
-  throttlingServiceProvider.get()
+  buildScan.background {
```
runningcode, line 33: @jprinet This seems odd that we need to do this. Is there a conversation with the configuration cache team about this?

If not, could you follow up with the configuration cache team to see if there is a better way to do this or a bug must be filed?

runningcode (approved):
looks good, thanks for the simplification!

jprinet:
This shall be the final version @runningcode.
Thanks again @abstratt and @mlopatkin for helping me 🙏

build-data-capturing-gradle-samples/capture-thermal-throttling/gradle-thermal-throttling.gradle
```diff
@@ -0,0 +1,149 @@
...
+      // process results on build completion
+      service.processResults(buildScanApi)
+    }
+  }
+} else {
+  println "INFO - Not running on MacOS - no thermal throttling data will be captured"
+}
+
+// Thermal Throttling service implementation
+abstract class ThermalThrottlingService implements BuildService<BuildServiceParameters.None>, AutoCloseable, OperationCompletionListener {
```
runningcode, line 50: Very interesting that we can use the `OperationCompletionListener` to mark the `BuildService` lifecycle to work with CC. Is there a bug filed somewhere for this?

build-data-capturing-gradle-samples/capture-thermal-throttling/gradle-thermal-throttling.gradle
```diff
@@ -0,0 +1,149 @@
...
+    scheduler.scheduleAtFixedRate(new ProcessRunner(COMMAND_ARGS, this::processCommandOutput), 0, SAMPLING_INTERVAL_IN_SECONDS, TimeUnit.SECONDS)
+  }
+
+  @Override
+  void close() {
+    scheduler.shutdownNow()
+  }
+
+  @Override
+  void onFinish(FinishEvent var1){
```
runningcode, line 85:
```suggestion
  void onFinish(FinishEvent ignored){
```

runningcode (approved):
Looking good, thanks for the final round of simplification!

build-data-capturing-gradle-samples/capture-thermal-throttling/gradle-thermal-throttling.gradle
jprinet: This is the way it works afaik, I leave it to @mlopatkin to confirm

mlopatkin: Revival of registered `OperationCompletionListener`s is a supported feature, though the intended use case is for actual listeners. Anyway, currently we don't have anything better for service bootstrapping and this API isn't going away soon, so using it looks fine to me.

PR #466, opened 2022-09-28
Configuration Cache compatible version of the Capture thermal throttling sample

jprinet:
I initially created [this PR](https://github.com/gradle/gradle-enterprise-build-config-samples/pull/461) to capture thermal throttling.
Using `buildFinished` hook made it incompatible with configuration cache.

Inspired by [this](https://github.com/gradle/gradle/issues/20151#issuecomment-1235358154), I adjusted the implementation which is now working fine when configuration cache is enabled


jprinet:
[Using GE plugin](https://docs.gradle.com/enterprise/gradle-plugin/#adding_data_at_the_end_of_the_build) is the way to go

PR #478, opened 2022-10-14
Specify GE version requirement for build cache auth snippet

alextu:


common-gradle-enterprise-gradle-configuration/settings.gradle
```diff
@@ -22,6 +22,7 @@ buildCache {
     }
 
     // Use the Gradle Enterprise connector's access key based authentication
+    // starting from Gradle Enterprise 2022.3 and Gradle Enterprise Plugin 3.11.
```
erichaagdev, line 25: Does it make sense to add a similar comment to the non-connector approach? Or maybe extend the comment to say "otherwise use the build cache configuration below".

erichaagdev (approved):

PR #501, opened 2022-11-17
New project common-gradle-enterprise-gradle-configuration using kts

cdsap:
New sample common-gradle-enterprise-gradle-configuration using kts


.github/dependabot.yml
```diff
@@ -22,6 +22,10 @@ updates:
     directory: "common-gradle-enterprise-gradle-configuration"
```
runningcode, line 22: i think we should rename this one to `-groovy` in order to prevent confusion.

common-gradle-enterprise-gradle-configuration-kts/README.md
```diff
@@ -0,0 +1,3 @@
+## Common Gradle Enterprise Gradle configuration with kts
```
runningcode, line 1:
```suggestion
## Common Gradle Enterprise Gradle configuration with KTS buildscripts
```

runningcode: let's also update the title in the groovy version :)

.github/dependabot.yml
cdsap: done!

PR #515, opened 2022-11-25
Fix workflow for Kotlin DSL capturing samples

erichaagdev:
I noticed the new workflow to verify the capturing samples using the Kotlin DSL subproject was referencing a `build.gradle` file instead of `build.gradle.kts`. It was also using Groovy DSL (`apply from: ...`) instead of Kotlin DSL (`apply(from = ...)`) to apply the capturing script.

I also update the steps to use the `working-directory` property to make it consistent with the surrounding steps.

### DIFF
--- a/.github/workflows/gradle-data-capturing-samples-verification.yml
+++ b/.github/workflows/gradle-data-capturing-samples-verification.yml
@@ -40,18 +40,20 @@ jobs:
       - name: Set up Gradle
         uses: gradle/gradle-build-action@v2
       - name: Inject data capture script into Gradle build using Groovy DSL
+        working-directory: common-gradle-enterprise-gradle-configuration-groovy
         run: |
           # apply sample file
-          echo "apply from: file(\"../build-data-capturing-gradle-samples/${{matrix.sample-file}}\")" >> common-gradle-enterprise-gradle-configuration-groovy/build.gradle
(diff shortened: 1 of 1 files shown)


runningcode (approved):
nice catch!

i'm wondering how this passed if that file does not exist.

cdsap (approved):

erichaagdev:
Good question.

Here is [a Build Scan comparison](https://ge.solutions-team.gradle.com/c/nu7ttq7dvmjfs/xhhjmjnyov2nm/custom-values?display-identical#ge-gradle-plugin-version-2) between `groovy` and `kotlin` for a build without these changes. 

In both builds, you can see the Gradle Enterprise plugin version did apply, so the verification did "work". However, the `kotlin` DSL run executed using the `build.gradle` file, not the `build.gradle.kts` file.

erichaagdev:
For completeness, here is [a Build Scan comparison](https://ge.solutions-team.gradle.com/c/n6nh6j3plb5le/2hetdf2qb5jd6/custom-values?display-identical#ge-gradle-plugin-version-2) between two builds with these changes which shows the `kotlin` DSL file is being properly updated.

runningcode:
Ah, i see now, the groovy projected was being executed and not the kotlin one as well! Thanks for fixing that!

PR #517, opened 2022-11-29
Remove unecessary task.configure block in quality check sample

alextu:
Works fine without and does not cause eager task configuration


etiennestuder:
@alextu Thanks. Please merge.

PR #544, opened 2022-12-21
Update java17 gha

cdsap:


.github/workflows/gradle-data-capturing-samples-verification.yml
```diff
@@ -30,10 +30,10 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v3
-      - name: Set up JDK 11
+      - name: Set up JDK 17
         uses: actions/setup-java@v3
         with:
-          java-version: '11'
+          java-version: '17'
           distribution: 'adopt'
```
erichaagdev, line 37: We should update these to use the `temurin` distribution. 

https://github.com/marketplace/actions/setup-java-jdk#supported-distributions

cdsap: Sorry Eric, but I'm missing the reason. Why do we need to update to temurin?

cdsap: I get it:
```
NOTE: AdoptOpenJDK got moved to Eclipse Temurin and won't be updated anymore. It is highly recommended to migrate workflows from adopt to temurin to keep receiving software and security updates. See more details in the [Good-bye AdoptOpenJDK post](https://blog.adoptopenjdk.net/2021/08/goodbye-adoptopenjdk-hello-adoptium/).
```

PR #550, opened 2022-12-23
Remove Findbugs from sample.

runningcode:
It is no longer supported. https://docs.gradle.org/current/userguide/upgrading_version_5.html#the_findbugs_plugin_has_been_removed


erichaagdev:
Should we remove Findbugs from the Maven sample as well?

runningcode:
I think that's a good idea too. It looks like the maven plugin was archived : https://github.com/gleclaire/findbugs-maven-plugin

erichaagdev (approved):
🔎 🪲 🚮

PR #559, opened 2023-01-05
Update AspectJ build caching configuration

guylabs:
The additionalAspectPaths property needs to be tracked as file set and only on the test compile goal, as the compile goal returns null for that property.


runningcode (approved):
Thanks for keeping this updated.

PR #603, opened 2023-02-24
Adding sample for Spotless Maven Plugin

cdsap:
Adding caching sample for the Spotless Maven Plugin (#604).

Plugin: https://github.com/diffplug/spotless/tree/main/plugin-maven
Main Mojo class: https://github.com/diffplug/spotless/blob/main/plugin-maven/src/main/java/com/diffplug/spotless/maven/AbstractSpotlessMojo.java

The configuration applies for the goals `check` and `apply`

****Tests****
**`apply` goal** 
In a project with an incorrect Java format like:
```
    System.out.println("Please" + "fix" +
        " me :)");
```
we'll execute: 
```
mvn clean spotless:apply
```
with the result:
Screenshot 2023-02-24 at 10 46 25 AM
https://ge.solutions-team.gradle.com/s/j224ntybqrxyw

The content of the file will be formatted like:
```
System.out.println("Please" + "fix" + " me :)");
```

If we apply again the goal `apply` the outcome is:
Screenshot 2023-02-24 at 10 48 13 AM
https://ge.solutions-team.gradle.com/s/uzak6wjk3yugu 

**`check` goal**

Second time we apply the goal `check` the result is:
Screenshot 2023-02-24 at 10 49 56 AM
https://ge.solutions-team.gradle.com/s/goxh3y37m65ci/timeline


cdsap:
@gradle/ge-solutions I would like to test Remote Cache for this sample, suggestions?

jprinet:
> @gradle/ge-solutions I would like to test Remote Cache for this sample, suggestions?

If not mistaken, you already have it checked by t[his GH workflow](https://github.com/gradle/gradle-enterprise-build-config-samples/blob/e41fdbd5065ba33185bea628751b1271418c0ede/.github/workflows/maven-build-caching-samples-verification.yml#L24):
https://github.com/gradle/gradle-enterprise-build-config-samples/actions/runs/4265191815/jobs/7424230699#step:6:4649

cdsap:
@runningcode I wanted to test both caches in these scenarios, my tests were executed with local cache, I wanted to verify both caches in case of missing some paths related issues in one of the inputs added. 

Good point about `apply`, @bigdaz mentioned too in the pre-analysis. Anyways, I understand that changes in-place causes non-reproducible builds, my idea was to cover the `apply` goal when is executed always in the verification phase, the reason behind is that we are not going to apply the format for the same file because always happens on the first execution. That's why we could produce cacheable outcomes for this goal on the following executions

cdsap:
@jprinet @runningcode I applied the changes

build-caching-maven-samples/spotless-project/pom.xml
```diff
@@ -0,0 +1,183 @@
...
+                                            </property>
+                                            <property>
+                                                <name>upToDateChecking</name>
+                                            </property>
+                                        </properties>
+                                        <ignoredProperties>
+                                            <ignore>repositorySystemSession</ignore>
+                                            <ignore>repositories</ignore>
+                                            <ignore>project</ignore>
+                                            <ignore>lineEndings</ignore>
```
runningcode, line 99: Do you know if this property would affect the outcome of the `check` goal?

build-caching-maven-samples/spotless-project/pom.xml
```diff
@@ -0,0 +1,183 @@
...
+                                            <name>pom</name>
+                                        </property>
+                                        <property>
+                                            <name>markdown</name>
+                                        </property>
+                                    </nestedProperties>
+                                    <outputs>
+                                        <directories>
+                                            <directory>
+                                                <name>buildDir</name>
```
runningcode, line 169: i wonder if there are any other outputs to cache, like reports? or are those all in here?

runningcode (approved):
just a few comments, otherwise looks good to me.

build-caching-maven-samples/spotless-project/pom.xml
cdsap: I'm not sure if is going to be affected

build-caching-maven-samples/spotless-project/pom.xml
cdsap: I verified again and I only saw https://github.com/diffplug/spotless/blob/main/plugin-maven/src/main/java/com/diffplug/spotless/maven/AbstractSpotlessMojo.java#L352

cdsap:
merging!

PR #609, opened 2023-03-01
Dependabot runs at a consistent time each day (02:00 UTC)

erichaagdev:
This PR sets the time that Dependabot will check for new dependencies to 02:00 UTC. This is aligned with the time that the wrapper upgrade runs. The idea is to have all version checks occur at the same time each day across all Solutions projects. 

https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#scheduletime

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -10,34 +10,40 @@ updates:
     directory: "build-caching-maven-samples"
     schedule:
       interval: "daily"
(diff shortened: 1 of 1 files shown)


erichaagdev:
I changed the branch name and did not mean to close this PR. 

Please see the new PR here #610

PR #611, opened 2023-03-01
Fix duplicate CI jobs from running on PRs

erichaagdev:
You may notice on other PRs that we are running each job twice. For example:

image

Notice the duplicate job names. This is because they are matching both the `pull_request` and `push` condition of the workflow.

This PR updates the conditions for when CI jobs are triggered in such a way that there is no overlap. Jobs will run when either:

- A pull request against `main` is opened or reopened or when the head branch of the pull request is updated, or
- A push to the `main` branch has been made

⚠️ This does create a gap between current behavior in that with these changes CI will not run for branches which do not have an open pull request or are not `main`. Additionally, CI will not run if an open pull request is not pointing against `main`.

I am open to ideas on other solutions. This is what is already being done in [android-cache-fix-gradle-plugin](https://github.com/gradle/android-cache-fix-gradle-plugin/blob/631b36ea45d9a9f40280b0719a1e9d5e68f00d91/.github/workflows/build-verification.yml#L3-L8) and is also mentioned [as a recommendation here](https://github.com/orgs/community/discussions/26276).

Once we come to a conclusion, I can update our other repositories where this is happening too. This one just suffers the most from this issue.

### DIFF
--- a/.github/workflows/gradle-data-capturing-samples-verification.yml
+++ b/.github/workflows/gradle-data-capturing-samples-verification.yml
@@ -1,6 +1,11 @@
 name: Verify Gradle Data Capturing Samples
 
-on: [ push, pull_request, workflow_dispatch ]
+on:
+  push:
+    branches: [ main ]
+  pull_request:
+    branches: [ main ]
+  workflow_dispatch:
 
 jobs:
   verification:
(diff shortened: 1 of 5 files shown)


runningcode (approved):

erichaagdev:
We discussed this internally and feel the changes provided by this PR is worth the change in functionality on when CI runs.

PR #612, opened 2023-03-02
Best effort fix for CodeNarc caputring issue

britter:
The previous version of the code assumed that the CodeNarc XML report
would always just contain a single `SourceDirectory` element. It turns
out that that is not true with the CodeNarc version we're using. There's
a `SourceDirectory` element for each package of the tested module. The
code then prepended all the `SourceDirectory` elements in front of each
file, which resulted in broken path being rendered into the custom
value.

With this commit the code now tries to find the right `SourceDirectory`
element by iterating over all of them and then checking whether the
violating file exists in that directory. This results in correct paths
being rendered. However it can result in the wrong file being reported
in case files with identical file names exist in different directories.


britter:
This is just updating the Groovy version of the script. The Kotlin version needs to have the same changes.

jprinet:
Thanks for the PR @britter, it is is indeed fixing the multi source directory case and not harming the default case.
Would it be possible for you to create the Kotlin version as well?

clayburn (approved):

PR #738, opened 2023-06-01
User count script

runningcode:
This adds a script to count the active committers to a series of git repositories in the last 30 days.
The series of repositories are listed in a `repositories.txt` file.

See `README.md` for how to use the script.

See specification document for more information: https://docs.google.com/document/d/18srJrOtvJUZpGYyneivb3bKO4c0L0vL0GQLzPjBjP3g/edit#


git-user-count/README.md
```diff
@@ -0,0 +1,24 @@
...
+
+### Overview
+
+The Gradle Enterprise Git user count script provides a means to automate the counting of the number of unique active developers in separate Git repositories.
+
+### Usage
+
+1. Update the `repositories.txt` file with the list of Git repositories that are connected to Gradle Enterprise.
+   Make sure each line contains a single Git repository URL.
+2. Run the `./rollout.sh` bash script to count the number of users rollout. The script supports the following command line arguments:
```
clayburn, line 11: I believe this is `user-count.sh` now

git-user-count/user-count.sh
```diff
@@ -0,0 +1,138 @@
...
+  sort -u "$userListFile" > "$userUniqueListFile"
+}
+
+function process_repository() {
+  repository_name="${1//\//-}"
+  
+  # Check if repository is valid
+  if [ -z "$repository_name" ]; then
+    echo "Repository name is empty" >&2
+    exit 1
```
clayburn, line 75: Is exiting the entire script necessary here? The result seems to be that you get:
* The CSV file for all processed repos until the failure
* The non-unique users for every repo processed so far
* An empty unique users file

These files give the impression that the script has completed, even though it has not. It seems that maybe just returning early from this function would just skip the empty line?

git-user-count/user-count.sh
```diff
@@ -0,0 +1,138 @@
...
+  popd >& /dev/null || return
+}
+
+function print_results() {
+  echo ""
+  echo "All cloned repositories available at $checkout_area"
+  echo "Unique usernames are stored in $(basename "$userUniqueListFile")"
+  echo "User counts by repository are stored in $(basename "$usersByRepoFile")"
+
+  echo "Total number of unique users: $(wc -l < "$userUniqueListFile")"
```
clayburn, line 119: This number, which is the main output of the script, doesn't standout visually. I wonder if an empty line above it would help, much link the empty line in the implementation of this function.

git-user-count/user-count.sh
```diff
@@ -0,0 +1,138 @@
+#!/bin/bash
+set -e
+
+basedir="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
+repositories=$basedir/repositories.txt
```
jprinet, line 5: There are no surrounding quotes for the value

git-user-count/user-count.sh
```diff
@@ -0,0 +1,138 @@
...
+userListFile="$basedir/gradle-enterprise-users.txt"
+userUniqueListFile="$basedir/gradle-enterprise-unique-users.txt"
```
jprinet, line 7: This is minor but it would be nice if the variable name could explicit if the file is an input or an output

git-user-count/user-count.sh
```diff
@@ -0,0 +1,138 @@
...
+  echo "All cloned repositories available at $checkout_area"
+  echo "Unique usernames are stored in $(basename "$userUniqueListFile")"
+  echo "User counts by repository are stored in $(basename "$usersByRepoFile")"
+
+  echo "Total number of unique users: $(wc -l < "$userUniqueListFile")"
+  echo "Note: this user count may include bots and other non-human users that commit to repositories."
+}
+
+function cleanup() {
+  rm "$userListFile"
```
jprinet, line 124: I would either remove all outputs (including `userUniqueListFile`) or none

jprinet (approved):
LGTM

git-user-count/user-count.sh
runningcode: I specifically remove just this one since it is confusing to have a list of users that is non unique. The unique user list is meant to be examined so that you can remove duplicates and CI users (as explained in the README and script output).

Let me know if I can make that more clear.

git-user-count/user-count.sh
runningcode: Good point.

runningcode: I removed the exit. let me know what you think.

git-user-count/user-count.sh
runningcode: added spacing before and after.

git-user-count/user-count.sh
runningcode: i think this is a good point. i added comments to clarify this in the script.

git-user-count/user-count.sh
clayburn: I think the exit needs be replaced with a `return 0` rather than omitted entirely for this to work. Otherwise, the, the git commands at the end of the function run from the `.tmp/repos` directory. This could have a few undesired outcomes:
- If this is running within a git repository, as I'm doing in build config samples here, the git commands run on that repository. So it counts committers to that outer repo and also runs a `git reset HEAD .` on that repo.
- If it is running outside of a git repo, then I would expect the git commands to error.

git-user-count/user-count.sh
```diff
@@ -0,0 +1,144 @@
+#!/bin/bash
```
jthurne, line 1: This is slightly more portable (and is a best practice):

```suggestion
#!/usr/bin/env bash
```

git-user-count/user-count.sh
```diff
@@ -0,0 +1,144 @@
...
+set -e
+
+basedir="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
```
jthurne, line 4: This is slightly more robust and deals better with symbolic links and the like:

```suggestion
basedir="$(cd "$(dirname "${BASH_SOURCE[0]}")"; cd -P "$(dirname "$(readlink "${BASH_SOURCE[0]}" || echo .)")"; pwd)"
```

git-user-count/user-count.sh
```diff
@@ -0,0 +1,144 @@
...
+    s) since=${OPTARG};;
+    c) shallow_clone=false;;
+    o) git_options=${OPTARG};;
+    *) echo "Usage: $0 [-b branch_name] [-s since_days] [-o git_options] [-c]" >&2
+       exit 1;;
+  esac
+done
+
+# Check that both -s and -c cannot be specified at the same time
+if [ "$shallow_clone" = false ] && [ -n "$since" ]; then
```
clayburn, line 40: I can't get past this check when I pass `-c`. I think this is because `since` is defaulted to 30, but doesn't get cleared out if `-c` is passed.

git-user-count/user-count.sh
```diff
@@ -0,0 +1,144 @@
+#!/bin/bash
+set -e
+
+basedir="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
+
+# input
+repositories="$basedir/repositories.txt"
```
jthurne, line 7: I think it would be a little more usable to take this as an CLI argument. It's okay if there is a default, but letting the user specify it will allow them to more easily specify different files for different uses. For example, maybe a company wants to know how many users there are by department, so they have different files for each department.

git-user-count/user-count.sh
jthurne: Probably the easiest way to fix this would be to test to check if `since` contains the default value (instead of checking to see if it is set).

Alternatively, the default could be assigned after this check runs.

git-user-count/user-count.sh
```diff
@@ -0,0 +1,144 @@
...
+
+function process_repository() {
+  repository_name="${1//\//-}"
+  
+  # Check if repository is valid
+  if [ -z "$repository_name" ]; then
+    echo "repository name is empty" >&2
+  # If repository is already cloned, switch to specified branch
+  elif [ -d "$checkout_area/$repository_name" ]; then
+    echo "already cloned."
```
jthurne, line 83: It might be a good idea to update the existing clone:

```
git pull
```

Otherwise, the script will only find data from when the clone was first created.

git-user-count/user-count.sh
```diff
@@ -0,0 +1,144 @@
...
+    if [ -n "$branch" ]; then
+      git_clone_command="${git_clone_command} --branch $branch"
+    fi
+    
+    # If shallow clone is enabled, append it to the git clone command
+    if [ "$shallow_clone" = true ]; then
+      git_clone_command="${git_clone_command} --shallow-since=\"${since} days ago\""
+    fi
+
+    eval "${git_clone_command}" && echo -e "${yellow}done${nc}"
```
jthurne, line 102: It would be slightly safer to not use eval here, and instead leave the `git clone` part out of the command string. It would look something like this:

```
git clone "${git_clone_args}" && echo -e "${yellow}done${nc}"
```

git-user-count/user-count.sh
```diff
@@ -0,0 +1,144 @@
...
+    git_clone_command="git clone --no-checkout  --filter=tree:0 --single-branch --no-tags $1 $checkout_area/$repository_name $git_options --quiet"
+
+    # If branch is specified, append it to the git clone command
+    if [ -n "$branch" ]; then
+      git_clone_command="${git_clone_command} --branch $branch"
+    fi
+    
+    # If shallow clone is enabled, append it to the git clone command
+    if [ "$shallow_clone" = true ]; then
+      git_clone_command="${git_clone_command} --shallow-since=\"${since} days ago\""
```
jthurne, line 99: The script seems to use varying styles (brackets or no brackets) when referring to variables. It would be good to standardize on one:

  1. `"$since"`
  2. `"${since}"`

git-user-count/user-count.sh
```diff
@@ -0,0 +1,144 @@
...
+    o) git_options=${OPTARG};;
+    *) echo "Usage: $0 [-b branch_name] [-s since_days] [-o git_options] [-c]" >&2
+       exit 1;;
+  esac
+done
+
+# Check that both -s and -c cannot be specified at the same time
+if [ "$shallow_clone" = false ] && [ -n "$since" ]; then
+  echo "Cannot specify both -s and -c options at the same time" >&2
+  exit 1
```
jthurne, line 42: I suggest printing the usage here as well. Maybe extract printing the usage to a function.

git-user-count/user-count.sh
```diff
@@ -0,0 +1,144 @@
...
+  echo ""
+  echo "Note: this user count may include bots and other non-human users that commit to repositories. Please examine the contents of $(basename "$userUniqueListFile") to remove duplicated and non-human users."
+}
+
+function cleanup() {
+  rm "$userListFile"
+}
+
+# entry point
+if [ ! -f "repositories.txt" ]; then
```
jthurne, line 134:
```suggestion
if [ ! -f "${repositories}" ]; then
```

git-user-count/user-count.sh
```diff
@@ -0,0 +1,144 @@
...
+  # If repository is already cloned, switch to specified branch
+  elif [ -d "$checkout_area/$repository_name" ]; then
+    echo "already cloned."
+    if [ -n "$branch" ]; then
+      pushd "$checkout_area/$repository_name" >& /dev/null || return
+      git checkout "$branch"
+      popd >& /dev/null || return
+    fi
+  else
+    git_clone_command="git clone --no-checkout  --filter=tree:0 --single-branch --no-tags $1 $checkout_area/$repository_name $git_options --quiet"
```
jthurne, line 90: I think how this command is built is probably fine the way it is. You might run into some shell quoting issues and the like with `git_options`. If you do, you could instead build up the command using  a bash array.  See [the `invoke_gradle` function from the Build Validation Scripts](https://github.com/gradle/gradle-enterprise-build-validation-scripts/blob/main/components/scripts/lib/gradle.sh) for an example.

git-user-count/user-count.sh
jprinet: I think an option to decide on the behavior could make sense, the current implementation allows to resume from a previous run. This can be very helpful when encountering an error with a large list of repos.

git-user-count/user-count.sh
runningcode: Thanks for checking this Clay, seems like I mistakenly refactored this bit of code.
On second pass I will remove this check.
I was originally thinking that the `since` doesn't make sense since when not using a shallow clone but the `since` is also used when calling `git log --since` so it makes sense to use both options at the same time.

git-user-count/user-count.sh
runningcode: Thanks.

git-user-count/user-count.sh
runningcode: These are both good points. This logic was originally intented to be used as Jerome wrote - to resume from a previous run. I would think this is the majority of the use cases here. I don't imagine that someone would keep all these repos around and then run them months later.

git-user-count/user-count.sh
runningcode: Good point. I cleaned this up and standardized on 1 except in cases where it is not possible to use that syntax.

git-user-count/user-count.sh
runningcode: I couldn't get this to work. it changed the way the arguments were parsed and quoted.
It would then say 
```
error: unknown option `no-checkout --filter=tree:0 --single-branch --no-tags https://github.com/gradle/maven-build-scan-quickstart .tmp/repos/https:--github.com-gradle-maven-build-scan-quickstart  --quiet --shallow-since="30 days ago"'
```

Is there another line I need to change to get this to work?

git-user-count/user-count.sh
runningcode: I thought that was a bit of an overkill, I gave it options with and without quotes as well as multiple options and it seemed to work. Do you know any specific cases I should try?

git-user-count/user-count.sh
jthurne: There are a number of ways to possibly fix this, but the simplest would be to switch to using an array to build the arguments.

That said, in this particular case, the use of `eval` isn't terrible, just not ideal. It's fine with me if you want to keep this the way it is for now.

runningcode: ok! we can always improve on this later

PR #831, opened 2023-08-22
Add Maven build caching sample for jacoco

tylerbertrand:
Adds sample project enabling caching for `jacoco:report` and `jacoco:aggregate-report` goals.

For both goals, retrieving the generated coverage reports from cache is faster than executing the goal with caching disabled as well as executing the goal with caching enabled and storing the cache artifact.

One consideration is the size of the cache artifact. Using Apache Dubbo as a reference, [the cache artifact for just one of its subprojects is 4.7MiB](https://ge.solutions-team.gradle.com/s/kjdhx37w5fhay/timeline?details=vnssj7leydar4&expanded=WyIwIl0&toggled=WyIxMDAiXQ&view=by-type). This could result in increased cache size, which may not be desirable. The [JacocoReport Gradle task is cacheable](https://github.com/gradle/gradle/blob/master/subprojects/jacoco/src/main/java/org/gradle/testing/jacoco/tasks/JacocoReport.java), which is worth noting.

### jacoco:report

* [Apache Dubbo ReportMojo, caching disabled: 4.5s serial execution](https://ge.solutions-team.gradle.com/s/a5aza5wjcz2xw/timeline?view=by-type#660)
* [Apache Dubbo ReportMojo, caching enabled, miss, store: 5.7s](https://ge.solutions-team.gradle.com/s/4tuoanthq3efu/timeline?view=by-type#533)
* [Apache Dubbo ReportMojo, caching enabled, hit: 0.23s serial execution](https://ge.solutions-team.gradle.com/s/42pi7geawnn56/timeline?view=by-type#1665)

### jacoco:report-aggregate

* [Apache Dubbo ReportAggregateMojo, caching disabled: 1m 4s serial execution](https://ge.solutions-team.gradle.com/s/36d2apjfrls6u/timeline?view=by-type#100)
* [Apache Dubbo ReportAggregateMojo, caching enabled, miss, store: 1m 32s serial execution](https://ge.solutions-team.gradle.com/s/kjdhx37w5fhay/timeline?expanded=WyIwIl0&view=by-type#100)
* [Apache Dubbo ReportAggregateMojo, caching enabled, hit: 36s serial execution](https://ge.solutions-team.gradle.com/s/qyc4vaxd7qvuk/timeline?view=by-type#100)


cdsap (approved):

etiennestuder:
@tylerbertrand Just for my understanding, since I never followed our Jacoco support closely: is what we support out of the box and document in the extension user manual non-sufficient, or does this sample cover another use case?

build-caching-maven-samples/jacoco-project/pom.xml
```diff
@@ -0,0 +1,140 @@
...
+                                                    <name>project</name>
+                                                    <nestedProperties>
+                                                        <property>
+                                                            <name>artifactId</name>
+                                                        </property>
+                                                        <property>
+                                                            <name>build</name>
+                                                            <nestedProperties>
+                                                                <property>
+                                                                    <name>outputDirectory</name>
```
runningcode, line 110: does this mean the outputDirectory is both an input and an output?

runningcode:
[Our docs](https://docs.gradle.com/enterprise/maven-extension/#jacoco_maven_plugin) state that none of the jacoco goals are cached out of the box and we instead recommend having the jacoco plugin hook in to the maven surefire and failsafe goals.

@tylerbertrand do you know if there is a use case here where it doesn't make sense to have the jacoco plugin hook in to the maven surefire and failsafe goals?

runningcode (changes requested):

build-caching-maven-samples/jacoco-project/pom.xml
tylerbertrand: So this `outputDirectory` is under the [MavenProject](https://maven.apache.org/ref/3.9.2/apidocs/org/apache/maven/project/MavenProject.html) parameter in jacoco's [AbstractReportMojo](https://github.com/jacoco/jacoco/blob/master/jacoco-maven-plugin/src/org/jacoco/maven/AbstractReportMojo.java#L99). Jacoco [does some processing on this output directory](https://github.com/jacoco/jacoco/blob/master/jacoco-maven-plugin/src/org/jacoco/maven/ReportSupport.java#L158-L159), so I thought it correct to define it as an input.

The `outputDirectory` listed under `outputs` is jacoco's [ReportMojo's outputDirectory property](https://github.com/jacoco/jacoco/blob/master/jacoco-maven-plugin/src/org/jacoco/maven/ReportMojo.java#L40-L41) and is where the report is written to.

tylerbertrand:
@etiennestuder, like @runningcode mentioned, jacoco goals are not cacheable by default. This sample enables caching for the `jacoco:report` goal.

runningcode:
When the Jacoco plugin hooks in to the surefire and failsafe goals does it not produce a report file? What do the surefire/failsafe goals produce if not a report?

build-caching-maven-samples/jacoco-project/pom.xml
runningcode: If jacoco does some processing on the output directory then it is both an input and an output and cannot be cached since this means the goal is not reproducible.

tylerbertrand:
> I saw your slack message. So does this mean that when the jacoco plugin hooks in to the surefire/failsafe goals then those produce an execution data file and the report goal transforms that file in to a report?

Yep that's exactly how it works. The `jacoco:prepare-agent` goal configures the java agent for the surefile/failsafe goals so that they produce the `jacoco.exec` file, and the `jacoco:report` goal will use that to create the coverage report.

runningcode:
Thanks for the analysis. I would be curious how the benchmark was set up here. Other processes running on your computer like IDEs, Gradle builds, and zoom calls will impact build times.
There is clearly a cost to caching this task as you see for both `report` and `report-aggregate`. In the `report-aggregate`, it seems to add 28 seconds to the build which then saves 28 seconds. We don't expect cache hits for all goals so this trade-off might not be worth it.

My understanding is that the `report-aggregate` is mostly copying files around. Is that also the case for the `report` goal?
Can you see what the tasks are doing?

tylerbertrand:
> I would be curious how the benchmark was set up here. Other processes running on your computer like IDEs, Gradle builds, and zoom calls will impact build times.

@runningcode Yeah I would not expect them to be 100% precise - there were definitely other things going on. I'd still think they would be fine to get an idea of whether or not they are worth caching.

> My understanding is that the report-aggregate is mostly copying files around. Is that also the case for the report goal? Can you see what the tasks are doing?

Both goals appear to do more or less the same thing - generate a coverage report from the class files and execution data file (jacoco.exec). The `report-aggregate` goal loads execution data and class files from all project dependencies and creates a combined report from them, while the `report` goal creates a single-project coverage report.

The goals do compute coverage, but it looks like most of the work is reading and traversing the execution data file along with the class files to compute that coverage, then writing the report.

> We don't expect cache hits for all goals so this trade-off might not be worth it.

Yep I think it is probably not worth it. After thinking about it I would expect these to be re-executed frequently.

PR #844, opened 2023-09-07
Prepare Quarkus extension for publication to Maven Central

jprinet:


quarkus-build-caching-extension/pom.xml
```diff
@@ -3,20 +3,46 @@
     <modelVersion>4.0.0</modelVersion>
     <groupId>com.gradle</groupId>
     <artifactId>quarkus-build-caching-extension</artifactId>
-    <version>0.9</version>
+    <version>0.9-SNAPSHOT</version>
```
runningcode, line 6: why are we starting at 0.9? just curious

For maven builds does adding a `-SNAPSHOT` automatically upload to the sonatype snapshots repository?

runningcode (commented):
Just a small question otherwise looks good.

quarkus-build-caching-extension/pom.xml
jprinet: 0.9 was the chosen version to be deployed internally.
Maven central only accepts releases, the `-SNAPSHOT` suffix is required by [Maven release plugin](https://maven.apache.org/maven-release/maven-release-plugin/prepare-mojo.html)

runningcode (approved):

PR #932, opened 2023-11-02
Add kotest to supported engines in gradle-test-pts-support scripts

tylerbertrand:
Applying [kotlin PTS compatibility script plugin](https://ge.solutions-team.gradle.com/s/zq5ufp34tk7xe/performance/configuration?details=code-unit-gradle-test-pts-support.gradle.kts!:&openScriptsAndPlugins=WyJjb2RlLXVuaXQtZ3JhZGxlLXRlc3QtcHRzLXN1cHBvcnQuZ3JhZGxlLmt0cyJd) to project using kotest - ([custom values showing kotest engine 5.8.0 (latest version) is compatible](https://ge.solutions-team.gradle.com/s/zq5ufp34tk7xe/custom-values))

Applying [kotlin PTS compatibility script plugin](https://ge.solutions-team.gradle.com/s/m7ufnwa2ax7h6/performance/configuration?details=code-unit-gradle-test-pts-support.gradle.kts!:&openScriptsAndPlugins=WyJjb2RlLXVuaXQtZ3JhZGxlLXRlc3QtcHRzLXN1cHBvcnQuZ3JhZGxlLmt0cyJd) to project using kotest - ([custom values showing kotest engine 5.6.0 (min compatible version) is compatible](https://ge.solutions-team.gradle.com/s/m7ufnwa2ax7h6/custom-values))

Applying [kotlin PTS compatibility script plugin](https://ge.solutions-team.gradle.com/s/n4v5y7d6g7fle/performance/configuration?details=code-unit-gradle-test-pts-support.gradle.kts!:&openScriptsAndPlugins=WyJjb2RlLXVuaXQtZ3JhZGxlLXRlc3QtcHRzLXN1cHBvcnQuZ3JhZGxlLmt0cyJd) to project using kotest - ([custom values showing kotest engine 5.4.0 is NOT compatible](https://ge.solutions-team.gradle.com/s/n4v5y7d6g7fle/custom-values))

### DIFF
--- a/build-data-capturing-gradle-samples/capture-test-pts-support/gradle-test-pts-support.gradle
+++ b/build-data-capturing-gradle-samples/capture-test-pts-support/gradle-test-pts-support.gradle
@@ -3,7 +3,6 @@ import org.gradle.util.internal.VersionNumber
 import java.nio.charset.StandardCharsets
 import java.util.Collections
 import java.util.Optional
-import java.util.function.Predicate
 import java.util.jar.JarFile
 import java.util.stream.Stream
 import java.util.stream.Collectors
@@ -64,7 +63,7 @@ class Capture {
             Stream<String> engines = t.classpath.files.stream()
                 .filter { f -> f.name.endsWith('.jar') }
                 .filter { f -> supportedEngines.values().stream().anyMatch { e -> f.name.contains(e) } }
-                .filter(filterIncompatibleKotestVersions(t))
+                .filter { f -> isCompatibleVersion(f, t) }
                 .map { f -> findTestEngine(f) }
                 .flatMap { o -> o.isPresent() ? Stream.of(o.get()) : Stream.empty() }
 
@@ -91,20 +90,18 @@ class Capture {
         }
     }
 
-    Predicate<File> filterIncompatibleKotestVersions(Test t) {
-        return { jar ->
-            if (jar.name.contains("kotest-runner")) {
-                def kotestVersionString = jar.name.split("-")[jar.name.split("-").length - 1].replace(".jar", "")
-                def kotestVersion = VersionNumber.parse(kotestVersionString)
-                if (VersionNumber.UNKNOWN == kotestVersion) {
-                    logger.error("Unable to parse kotest version from file name ${jar.name}")
-                    buildScanApi.value("${t.identityPath}#unknownKotestVersion", "${jar.name}")
-                    return false
-                }
-                return VersionNumber.parse("5.6.0") <= kotestVersion
-            } else {
-                return true
+    boolean isCompatibleVersion(File f, Test t) {
+        if (f.name.contains("kotest-runner")) {
+            def kotestVersionString = f.name.split("-")[f.name.split("-").length - 1].replace(".jar", "")
+            def kotestVersion = VersionNumber.parse(kotestVersionString)
+            if (VersionNumber.UNKNOWN == kotestVersion) {
+                logger.error("Unable to parse kotest version from file name ${f.name}")
+                buildScanApi.value("${t.identityPath}#unknownKotestVersion", "${f.name}")
+                return false
             }
+            return VersionNumber.parse("5.6.0") <= kotestVersion
+        } else {
+            return true
         }
     }
 }
--- a/build-data-capturing-gradle-samples/capture-test-pts-support/gradle-test-pts-support.gradle
+++ b/build-data-capturing-gradle-samples/capture-test-pts-support/gradle-test-pts-support.gradle
@@ -32,7 +32,8 @@ class Capture {
         'org.spockframework.runtime.SpockEngine' : 'spock',
         'net.jqwik.engine.JqwikTestEngine' : 'jqwik',
         'com.tngtech.archunit.junit.ArchUnitTestEngine' : 'archunit',
-        'co.helmethair.scalatest.ScalatestEngine' : 'scalatest'
+        'co.helmethair.scalatest.ScalatestEngine' : 'scalatest',
+        'io.kotest.runner.junit.platform.KotestJunitPlatformTestEngine' : 'kotest-runner'
     ]
     private Logger logger
 
--- a/build-data-capturing-gradle-samples/capture-test-pts-support/gradle-test-pts-support.gradle
+++ b/build-data-capturing-gradle-samples/capture-test-pts-support/gradle-test-pts-support.gradle
@@ -1,7 +1,9 @@
 import com.gradle.scan.plugin.BuildScanExtension
+import org.gradle.util.internal.VersionNumber
 import java.nio.charset.StandardCharsets
 import java.util.Collections
 import java.util.Optional
+import java.util.function.Predicate
 import java.util.jar.JarFile
 import java.util.stream.Stream
 import java.util.stream.Collectors
@@ -15,11 +17,11 @@ def buildScanApi = project.extensions.findByName('buildScan')
 if (!buildScanApi) {
     return
 }
-def capture = new Capture(gradle.rootProject.logger)
+def capture = new Capture(buildScanApi, gradle.rootProject.logger)
 allprojects {
     tasks.withType(Test).configureEach { t ->
         doFirst {
-            capture.capturePts(t, buildScanApi)
+            capture.capturePts(t)
         }
     }
 }
@@ -36,12 +38,14 @@ class Capture {
         'io.kotest.runner.junit.platform.KotestJunitPlatformTestEngine' : 'kotest-runner'
     ]
     private Logger logger
+    private BuildScanExtension buildScanApi
 
-    Capture(Logger logger) {
+    Capture(BuildScanExtension buildScanApi, Logger logger) {
+        this.buildScanApi = buildScanApi
         this.logger = logger
     }
 
-    void capturePts(Test t, BuildScanExtension buildScanApi) {
+    void capturePts(Test t) {
         if (t.getTestFramework().getClass().getName() == 'org.gradle.api.internal.tasks.testing.junitplatform.JUnitPlatformTestFramework') {
             def engines = testEngines(t)
             buildScanApi.value("${t.identityPath}#engines", "${engines}")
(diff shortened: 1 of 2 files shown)


erichaagdev:
Only versions 5.6.0 and later of Kotest is supported. If an incompatible version is used, how will it be reported?

runningcode (approved):
Thanks for updating this!

runningcode (changes requested):
Sorry meant to request changes. Let's see if we can address Eric's comment.

tylerbertrand:
@erichaagdev @runningcode I updated the kotlin script to account for the fact that only kotest 5.6.0 and later is supported. It is kotest-specific to keep version parsing simple, and because kotest is the only version-constrained engine we support so far. Let me know your thoughts.  

As for whether and how the incompatible kotest engine version should be reported...that would require a bit of a rework of the scripts. Right now we're more or less just checking that jar names match a hardcoded list of compatible versions. I'm not exactly sure of a straightforward way we would track which jars are test engines and which are not outside of kotest, which we know is a test engine because it's in the list of compatible ones.

build-data-capturing-gradle-samples/capture-test-pts-support/gradle-test-pts-support.gradle.kts
```diff
@@ -86,3 +96,30 @@ class Capture(val logger: Logger) {
         }
     }
 }
+
+class KotestEngine(private val version: String) : Comparable<KotestEngine> {
```
runningcode, line 100: Can we use Gradle's `VersionNumber` class to do the version parsing/comparison here?

And maybe taking a step back, we want to compare `VersionNumber`s not `KotestEngine`s

build-data-capturing-gradle-samples/capture-test-pts-support/gradle-test-pts-support.gradle.kts
```diff
@@ -86,3 +96,30 @@ class Capture(val logger: Logger) {
...
+
+    constructor(jarFile: File) : this(jarFile.name.split("-")[jarFile.name.split("-").lastIndex].replace(".jar", ""))
```
runningcode, line 102: is the version always part of the jar file's name?

build-data-capturing-gradle-samples/capture-test-pts-support/gradle-test-pts-support.gradle.kts
tylerbertrand: Updated to use `VersionNumber` to do the parsing/comparison.

build-data-capturing-gradle-samples/capture-test-pts-support/gradle-test-pts-support.gradle.kts
tylerbertrand: It is

build-data-capturing-gradle-samples/capture-test-pts-support/gradle-test-pts-support.gradle.kts
```diff
@@ -56,6 +57,19 @@ class Capture(val logger: Logger) {
...
                 .filter { f -> supportedEngines.values.stream().anyMatch { e -> f.name.contains(e) } }
+                .filter { f ->
+                    if (f.name.contains("kotest-runner")) {
+                        val kotestVersionString = f.name.split("-")[f.name.split("-").lastIndex].replace(".jar", "")
+                        val kotestVersion = VersionNumber.parse(kotestVersionString)
+                        if (VersionNumber.UNKNOWN == kotestVersion) {
+                            logger.warn("Unable to parse kotest version from file name ${f.name}")
+                            return@filter false
+                        }
+                        VersionNumber.parse("5.6.0") <= kotestVersion
```
erichaagdev, line 68: According to the JavaDoc, this [VersionNumber](https://docs.gradle.org/current/javadoc/org/gradle/util/VersionNumber.html) will be removed in Gradle 9.0. 

There is an internal one [`org.gradle.util.internal.VersionNumber`](https://github.com/gradle/gradle/blob/e64caf03d0ceed251f83a3e90281a4d8c5dc5f7e/platforms/core-runtime/base-services/src/main/java/org/gradle/util/internal/VersionNumber.java#L36). Should we use that instead?

build-data-capturing-gradle-samples/capture-test-pts-support/gradle-test-pts-support.gradle
```diff
@@ -32,7 +32,8 @@ class Capture {
         'org.spockframework.runtime.SpockEngine' : 'spock',
```
erichaagdev, line 32: Will you also be making the same changes for Groovy?

build-data-capturing-gradle-samples/capture-test-pts-support/gradle-test-pts-support.gradle.kts
```diff
@@ -56,6 +57,19 @@ class Capture(val logger: Logger) {
         try {
             var engines = t.classpath.files.stream().filter { f -> f.name.endsWith(".jar") }
                 .filter { f -> supportedEngines.values.stream().anyMatch { e -> f.name.contains(e) } }
+                .filter { f ->
+                    if (f.name.contains("kotest-runner")) {
```
erichaagdev, line 61: I would extract all of this comparison logic to a function.

build-data-capturing-gradle-samples/capture-test-pts-support/gradle-test-pts-support.gradle.kts
tylerbertrand: Fixed

tylerbertrand: And yeah I guess to answer the question, I think it makes sense to use that one, since the external one will be removed soon-ish, though I'm not sure if we should completely refrain from using internal packages

build-data-capturing-gradle-samples/capture-test-pts-support/gradle-test-pts-support.gradle
tylerbertrand: Yep, just wanted to nail down the Kotlin implementation, then I'll make the same changes to the Groovy script

build-data-capturing-gradle-samples/capture-test-pts-support/gradle-test-pts-support.gradle.kts
tylerbertrand: Extracted the filter predicate to a property

build-data-capturing-gradle-samples/capture-test-pts-support/gradle-test-pts-support.gradle.kts
```diff
@@ -85,4 +88,18 @@ class Capture(val logger: Logger) {
...
         }
     }
+
+    private val filterIncompatibleKotestVersions = Predicate<File> { jar ->
+        if (jar.name.contains("kotest-runner")) {
+            val kotestVersionString = jar.name.split("-")[jar.name.split("-").lastIndex].replace(".jar", "")
+            val kotestVersion = VersionNumber.parse(kotestVersionString)
+            if (VersionNumber.UNKNOWN == kotestVersion) {
+                logger.warn("Unable to parse kotest version from file name ${jar.name}")
+                return@Predicate false
```
runningcode, line 98: This looks good, I would just make both return statements consistent. Either they both have the `return@Predicate` or neither.

build-data-capturing-gradle-samples/capture-test-pts-support/gradle-test-pts-support.gradle.kts
```diff
@@ -85,4 +88,18 @@ class Capture(val logger: Logger) {
                 }
         }
     }
+
+    private val filterIncompatibleKotestVersions = Predicate<File> { jar ->
+        if (jar.name.contains("kotest-runner")) {
+            val kotestVersionString = jar.name.split("-")[jar.name.split("-").lastIndex].replace(".jar", "")
+            val kotestVersion = VersionNumber.parse(kotestVersionString)
+            if (VersionNumber.UNKNOWN == kotestVersion) {
+                logger.warn("Unable to parse kotest version from file name ${jar.name}")
```
runningcode, line 97: im wondering if we should put this as `error` level or just directly throw an exception so we can fix the error. we could also add a custom value here as well to prevent just thinking it was incompatible.

runningcode (commented):
just had some more thoughts, otherwise looks good

build-data-capturing-gradle-samples/capture-test-pts-support/gradle-test-pts-support.gradle.kts
tylerbertrand: Good point, I think an exception or custom value (since we are looking at custom values, and maybe not logs, when determining compatibility) would be best so that there is less of a chance the issue is missed.

tylerbertrand: Pushed a commit adding a custom value and logging the error, required a bit of rework to make the `BuildScanExtension` and `Test` task available to the predicate - lmk if you think this is overkill

tylerbertrand: Here's an example of how this would look https://ge.solutions-team.gradle.com/s/2ukw7odlyyge4/custom-values#L2 (i set the version to unknown to force the custom value to be created, normally this 5.7.2 version would be compatible)

build-data-capturing-gradle-samples/capture-test-pts-support/gradle-test-pts-support.gradle
```diff
@@ -32,15 +34,18 @@ class Capture {
...
+    Capture(BuildScanExtension buildScanApi, Logger logger) {
+        this.buildScanApi = buildScanApi
         this.logger = logger
     }
 
-    void capturePts(Test t, BuildScanExtension buildScanApi) {
+    void capturePts(Test t) {
         if (t.getTestFramework().getClass().getName() == 'org.gradle.api.internal.tasks.testing.junitplatform.JUnitPlatformTestFramework') {
             def engines = testEngines(t)
             buildScanApi.value("${t.identityPath}#engines", "${engines}")
```
runningcode, line 51: does this line still work?

runningcode (approved):

build-data-capturing-gradle-samples/capture-test-pts-support/gradle-test-pts-support.gradle.kts
runningcode: Thanks for the example, I think that will help in debugging!

erichaagdev (approved):
Nice!

PR #946, opened 2023-11-27
Add Gradle Enterprise SBT plugin configuration sample

tylerbertrand:
Adds a sample for common Gradle Enterprise SBT plugin configuration


common-gradle-enterprise-sbt-configuration/build.sbt
```diff
@@ -0,0 +1,11 @@
+// You can configure the Gradle Enterprise sbt plugin to publish Build Scans to your Gradle Enterprise server
+// by adding the following configuration to your project's build.sbt file
```
runningcode, line 2: This looks good, one thing missing here is setting the project name. Can you add that please? Then we'll send it to the sbt team for review.

tylerbertrand: Done

PR #1006, opened 2024-01-03
Added information about how to apply build capture samples for Gradle.

ribafish:


build-data-capturing-gradle-samples/README.md
```diff
@@ -7,15 +7,17 @@ a [script plugin](https://docs.gradle.org/current/userguide/plugins.html#sec:scr
...
 ```
 
 ```kotlin
-// build.gradle.kts
+// root build.gradle.kts
 apply(from = "build-data-capturing-sample.gradle.kts")
 ```
 
+The samples are designed to be applied to the root level `build.gradle(.kts)` script. If it's necessary to use them on a subproject `build.gradle(.kts)` script, the samples should be modified to reference the `rootProject` when retrieving the scan extension.
+
```
erichaagdev, line 20: The samples aren't designed to be applied to anything but the root project. It might be good to clarify this, but I don't think we should include the part about how to apply it to subprojects. 

Ultimately, these are just reference samples that should not be applied as script plugins long-term. Instead, they should be included in a convention plugin.

runningcode: I agree to remove the information on how to apply this to the subprojects. These are just samples.

ribafish: Removed

runningcode (approved):

PR #1007, opened 2024-01-03
M1 change to M1CHIP

rahat2134:
SInce it more about processor so  i tried to change the name from M1 to M1CHIP to make it more readable.


erichaagdev:
Hi @rahat2134. This change would break our GitHub Action workflows and does not actually make anything more readable. 

I am going to close this PR.

PR #1078, opened 2024-03-04
Replace Quarkus dependency checksum with Quarkus dependencies as goal input

jprinet:
### Context
Quarkus dynamically adds some dependencies to the build, which are listed in the `target/quarkus-prod-dependency-checksums.txt` file. 
This file contains one dependency per line and the fingerprint of the dependency for `SNAPSHOT` versions.
This file is then used as extra goal input for the Quarkus `build` goal and the `surefire` and `failsafe` test goals.

### Issue
The fingerprinting logic [is part of the Quarkus plugin](https://github.com/quarkusio/quarkus/pull/37622). 
This is something that would be naturally handled by the Develocity extension if the Quarkus plugin would only expose the list of files to fingerprint.
In addition, some optimizations could happen when computing the fingerprinting of a file already processed.
Eventually, delegating the fingerprinting to the Develocity Maven extension opens the door to normalization options.

### Fix
Collect the Quarkus extra dependencies from the `target/quarkus-prod-dependencies.txt` file exposed by the Quarkus Maven plugin and add them as file inputs with `RUNTIME_CLASSPATH` normalization strategy.

*Note:*
The logic to use `target/quarkus-prod-dependency-checksums.txt` as an input  is kept for compatibility reasons.

### Dependency
Modify the Quarkus Maven plugin implementation to create a `target/quarkus-prod-dependencies.txt` file containing absolute path to each of the dependencies, instead of the current `target/quarkus-prod-dependency-checksums.txt`

### DIFF
--- a/quarkus-build-caching-extension/README.md
+++ b/quarkus-build-caching-extension/README.md
@@ -8,6 +8,7 @@ This project is derived from [the CCUD Maven extension](https://github.com/gradl
 A native executable can be a very large file. Copying it from/to the local cache, or transferring it from/to the remote cache can be an expensive operation that has to be balanced with the duration of the work being avoided.
 
 ## Requirements
+- [Gradle Enterprise plugin](https://docs.gradle.com/enterprise/gradle-plugin/) until 3.16.x
 - Quarkus 3.2.4 and above which brings [track-config-changes goal](https://quarkus.io/guides/config-reference#tracking-build-time-configuration-changes-between-builds)
 
 *Note:*<br>


jprinet:
@aloubyansky @gsmet this change would allow to remove the fingerprinting logic from the Quarkus Maven plugin
wdyt?

aloubyansky:
We could do that. You could probably try prototyping it by simply ignoring the checksum values present in the current file, until we implement this on the Quarkus side.

jprinet:
> We could do that. You could probably try prototyping it by simply ignoring the checksum values present in the current file, until we implement this on the Quarkus side.

I actually need absolute paths, I don't want to magically rebuild the path to the Maven repository as custom configuration could lead to surprises.
If you agree with the approach, I could keep old and new implementation. This means both files would be processed as goal inputs, although only one of them would be present depending on the Quarkus version.

jprinet:
I pushed a new version @aloubyansky to support both `quarkus-prod-dependencies.txt` and `quarkus-prod-dependency-checksums.txt`.
Please let me know if you are ok with the changes (here and in the Quarkus Maven plugin)

PR #1121, opened 2024-04-05
Update to use DV Maven extension 1.21 and DV API

runningcode:
This also fixes the maven build caching samples, the maven data capturing samples, the maven rollout script, the creating a maven distribution training.

TODO: I still haven't figured out how to fix the capture develocity maven extension version script.


clayburn:
Regarding the failing Maven Build Caching Sample: https://ge.solutions-team.gradle.com/c/n26q3qjmlrujs/ad4vg3vhutpw4/goal-inputs?cacheability=cacheable

I've seen this `builddef.lst` miss before. [This](https://github.com/mojohaus/aspectj-maven-plugin/issues/52) is the issue from the aspectj plugin, which is not resolved. That `builddef.lst` file I believe is just an output of the `AjcCompileMojo` that contains the configuration of that compilation. A goal like surefire could safely ignore that file with normalization I believe.

The real question to me though is what about this change causes this to occur? My best guess is that it is the `-e` workaround, since it will inject the legacy Gradle Enterprise Maven Extension 1.18.1. I tried this locally (manually) using the Develocity Maven Extension 1.21 and could not reproduce the cache miss: https://ge.solutions-team.gradle.com/c/riufudxr7zxtg/hv2kwnj2l2xwm/goal-inputs

It isn't clear to me why this is the only case that would break. I think we really need an updated Build Validation Scripts that does require this workaround to make the most sense of this.

PR #1133, opened 2024-04-10
Fixed PTS script if not all dependencies are resolvable

ribafish:
If not all dependencies/configurations are resolvable, currently the PTS support detection script throws an [error that fails the build](https://ge.solutions-team.gradle.com/s/byyxejysywnd4/failure#1), even if the build otherwise succeeds.

With these changes, the [build will still pass](https://ge.solutions-team.gradle.com/s/uov4lngy6meii), but show a [warning about it being unable to resolve some of the configurations](https://ge.solutions-team.gradle.com/s/uov4lngy6meii/console-log?page=1#L13).


runningcode (approved):
Thanks for the fix

PR #1186, opened 2024-05-14
Develocity Gradle plugin is removed from convention plugin sample

erichaagdev:
As it exists today, the user is prompted to publish the scan to `scans.gradle.com`. This is because the Develocity Gradle plugin is applied, but not configured.

This change removes the Develocity Gradle and CCUD plugins.

### DIFF
--- a/convention-develocity-gradle-plugin/plugins/gradle-2-through-4/settings.gradle
+++ b/convention-develocity-gradle-plugin/plugins/gradle-2-through-4/settings.gradle
@@ -1,6 +1,4 @@
 plugins {
-    id 'com.gradle.develocity' version '3.17.3'
-    id 'com.gradle.common-custom-user-data-gradle-plugin' version '2.0.1'
     id 'org.gradle.toolchains.foojay-resolver-convention' version '0.8.0'
 }
 
--- a/convention-develocity-gradle-plugin/plugins/gradle-5-or-newer/settings.gradle
+++ b/convention-develocity-gradle-plugin/plugins/gradle-5-or-newer/settings.gradle
@@ -1,6 +1,4 @@
 plugins {
-    id 'com.gradle.develocity' version '3.17.3'
-    id 'com.gradle.common-custom-user-data-gradle-plugin' version '2.0.1'
     id 'org.gradle.toolchains.foojay-resolver-convention' version '0.8.0'
 }
 


convention-develocity-gradle-plugin/plugins/gradle-2-through-4/settings.gradle
```diff
@@ -1,7 +1,29 @@
 plugins {
-    id 'com.gradle.develocity' version '3.17.3'
+    id 'com.gradle.develocity' version '3.17.3' // replace with this convention plugin once published
```
etiennestuder, line 2: Not sure what you mean with this comment. There is no plan to publish this convention plugin. It serves as a template that other can copy and adjust to their needs.

convention-develocity-gradle-plugin/plugins/gradle-2-through-4/settings.gradle
```diff
@@ -1,7 +1,29 @@
 plugins {
```
etiennestuder, line 1: I don't understand the motivation of this PR. This is an example of how to build a convention plugin. I don't see the point in adding the DV configuration to it so that the project itself publishes build scans.

etiennestuder (changes requested):
I think we should no integrate this PR.

etiennestuder:
@erichaagdev Instead of adding the DV configuration to the settings.file, I would remove the DV/CCUD plugin declarations. The reason being that this project is an example of how to build a convention plugin, not how to configure a project with DV.

erichaagdev:
> I don't understand the motivation of this PR.

@etiennestuder the motivation was that I ran the build and was prompted to agree to the Gradle ToS to publish a scan to `scans.gradle.com`. This is because the build had the Develocity Gradle plugin applied, but it was not configured. This didn't feel right. I felt either of these options would be improvements over that state:

1. Configure the Develocity Gradle plugin
2. Remove the Develocity Gradle plugin

I chose option 1 to encourage and remind users that this build should also be connected to Develocity once it is copied. I chose `develocity-samples.gradle.com` as this is the same hypothetical server the convention plugin configures. 

But, maybe a better option is to remove Develocity Gradle plugin and CCUD as you say. We could then add this to the `settings.gradle` file instead: 

```groovy
plugins {
    // CHANGE ME: Apply your convention plugin once published
    // plugin('com.myorg.convention-develocity-gradle-5-or-newer') version '1.0.0'
    ..
}
```

This is the same pattern we follow for the rest of this template:

https://github.com/gradle/develocity-build-config-samples/blob/82714f7550189d6867b2627a7feba9c7de57d1ab/convention-develocity-gradle-plugin/plugins/gradle-5-or-newer/src/main/java/com/myorg/ConventionDevelocityGradlePlugin.java#L60

What do you think about this?

etiennestuder:
> What do you think about this?

Yep. I fully agree with your option #2.

https://github.com/gradle/develocity-build-config-samples/pull/1186#issuecomment-2111869268

convention-develocity-gradle-plugin/plugins/gradle-2-through-4/settings.gradle
erichaagdev: Resolving this comment given the approach has changed.

PR #1188, opened 2024-05-15
Convention Develocity Gradle plugin configures Develocity Gradle plugin

erichaagdev:
This PR updates the Convention Develocity Gradle plugin sample to configure the Develocity Gradle plugin.

Resolves gradle/dv#34435


convention-develocity-gradle-plugin/plugins/gradle-5-or-newer/src/main/java/com/myorg/ConventionDevelocityGradlePlugin.java
```diff
@@ -69,15 +67,14 @@ private boolean containsPropertiesTask(StartParameter startParameter) {
                 || startParameter.getTaskNames().stream().anyMatch(it -> it.endsWith(":properties"));
     }
 
-    private void configureBuildScan(BuildScanExtension buildScan) {
+    private void configureBuildScan(BuildScanConfiguration buildScan) {
         // CHANGE ME: Apply your Build Scan configuration here
-        buildScan.publishAlways();
-        buildScan.setUploadInBackground(!isCi());
+        buildScan.getUploadInBackground().set(isCi());
```
erichaagdev, line 72: I really wanted to update `isCi()` to use the Provider API and return a `Provider<Boolean>` but it just isn't worth it until the build cache configuration is updated to use Provider API.

convention-develocity-gradle-plugin/plugins/gradle-5-or-newer/src/main/java/com/myorg/ConventionDevelocityGradlePlugin.java
```diff
@@ -43,24 +41,24 @@ public void apply(Object target) {
...
         settings.getPluginManager().apply(CommonCustomUserDataGradlePlugin.class);
-        configureGradleEnterprise(settings.getExtensions().getByType(GradleEnterpriseExtension.class), settings.getGradle().getStartParameter());
-        configureBuildCache(settings.getBuildCache(), settings.getExtensions().getByType(GradleEnterpriseExtension.class));
+        configureDevelocity(settings.getExtensions().getByType(DevelocityConfiguration.class), settings.getGradle().getStartParameter());
+        configureBuildCache(settings.getBuildCache(), settings.getExtensions().getByType(DevelocityConfiguration.class));
     }
 
     private void configureGradle5(Project project) {
-        project.getPluginManager().apply(BuildScanPlugin.class);
+        project.getPluginManager().apply(DevelocityPlugin.class);
```
erichaagdev, line 51: Our docs say the Develocity Gradle plugin is compatible with Gradle 5.x and higher as opposed to having to use the Build Scan plugin. I tested it using the sample projects and it works.

https://docs.gradle.com/develocity/gradle-plugin/current/#applying_the_plugin

clayburn (approved):
LGTM

.github/workflows/convention-develocity-gradle-plugin-verification.yml
```diff
@@ -0,0 +1,64 @@
...
+          path: ~/.m2/repository/com/myorg
+
+  verification:
+    name: Verify Example Build for Gradle ${{ format(matrix.gradle-version, ' ') }}
+    needs: [build]
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        gradle-version: ['5', '6{0}and{0}later'] # Gradle 4 and earlier examples are not tested as they would not succeed
```
erichaagdev, line 40: Because 3 and earlier does not apply CCUD, so it's not possible to override the server URL from the command line. See [here](https://github.com/gradle/develocity-build-config-samples/actions/runs/9116910759/job/25066613181#step:6:40).

And because 4 does not support publishing to authenticated servers. See [here](https://github.com/gradle/develocity-build-config-samples/actions/runs/9116910759/job/25066613424#step:6:43).

clayburn: I would recommend just leaving this useful info as a comment in the action somewhere

.github/workflows/convention-develocity-gradle-plugin-verification.yml
```diff
@@ -0,0 +1,64 @@
...
+        uses: actions/checkout@v4
+      - name: Set up JDK 8
+        uses: actions/setup-java@v4
+        with:
+          java-version: '8'
+          distribution: 'temurin'
+      - name: Set up Gradle
+        uses: gradle/actions/setup-gradle@v3
+      - name: Build with Gradle
+        working-directory: convention-develocity-gradle-plugin/plugins/gradle-5-or-newer
```
clayburn, line 25: Should we add a similar verification that at least builds the gradle <4 version? Even if we can't verify the functionality works, knowing it compiles is at least something

erichaagdev: Yes. Good point.

erichaagdev: Done in 4cdccf04e6f19c6998bf2e27b0e2b5b820855021

.github/workflows/convention-develocity-gradle-plugin-verification.yml
erichaagdev: Done in db7b3d1cd808c78584ea0eda802e95c5f8901d23

PR #1194, opened 2024-05-20
Add quarkus extension integration tests

jprinet:
### Changes
- The integration tests are run with the [maven-invoker-plugin](https://maven.apache.org/plugins/maven-invoker-plugin/)
- A test scenario is comprised of 6 builds (see `invoker.properties`) to validate the caching mechanism
- The scenario is tested with 2 versions of Quarkus
  - 3.8.4 (LTS)
  - Latest snapshot
- A weekly CI job ensures the extension works with Quarkus latest version

### Logging
This pull request also contains some logging improvements to ease troubleshooting.
In particular, the check for dump configuration presence to allow cacheability is done as first thing which allows to have:
`Quarkus previous configuration not found`
instead of
`Quarkus package type null is not cacheable`

### Side-effect
- Adding integration tests allowed to spot an issue when using a deprecated Quarkus key (see [this commit](https://github.com/gradle/develocity-build-config-samples/commit/5338dfd4d2db119b64b296c1e35cf2a2323a41bf) for details)


quarkus-build-caching-extension/src/it/postbuild.groovy
```diff
@@ -0,0 +1,38 @@
+// Utility functions
+String getContent(String fileName) {
+    File f = new File(basedir, fileName)
+    assert f.exists()
+    return f.text
+}
+
+void checkBuildCacheDisabled(String logFile) {
```
erichaagdev, line 8: There is some strange spacing in the `contains()` invocations of these 3 fuctions.

erichaagdev (approved):

.github/workflows/quarkus-build-caching-extension.yml
```diff
@@ -1,10 +1,15 @@
 name: Verify Quarkus Build Caching Extension
 
 on:
+  schedule:
+    # Every Sunday at 9.00am - scheduled to test against latest Quarkus version
+    - cron: "0 9 * * 0"
   push:
     branches: [ main ]
+    paths: [ 'quarkus-build-caching-extension/**' ]
```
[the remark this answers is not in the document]
jprinet: Why not, I felt this is really required here as the integration tests would take around 10 minutes to run, which would be frustrating when changing totally unrelated code

.github/workflows/quarkus-build-caching-extension.yml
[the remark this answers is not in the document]
jprinet: this `999-SNAPSHOT` version is being overwritten on each snapshot deployment. Your comment made me realized that there is a more straightforward solution. WIP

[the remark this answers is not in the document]
jprinet: I changed the approach, the Quarkus Snapshot repository is now configured in the `pom.xml` to allow fetching directly the latest development version, which is even better than running tests against latest release (we would be warned earlier about a breaking change).
The Dependabot approach here wouldn't work as snapshots are published several times a day.

quarkus-build-caching-extension/src/it/postbuild.groovy
runningcode, line 8: This is a nit but I would name all these methods in this file with `ensure` or `assert` because once this method passes we know the build cache is disabled. `check` implies a return value.
```suggestion
void ensureBuildCacheDisabled(String logFile) {
```

quarkus-build-caching-extension/src/main/java/com/gradle/QuarkusBuildCache.java
```diff
@@ -100,12 +100,15 @@ private void configureQuarkusBuildGoal(MojoMetadataProvider.Context context, Qua
             if (extensionConfiguration.isQuarkusCacheEnabled()) {
                 LOGGER.debug(QuarkusExtensionUtil.getLogMessage("Quarkus caching is enabled"));
 
-                // Load Quarkus properties for current build
+                // Load Quarkus properties for previous build
```
runningcode, line 103:
```suggestion
                // Load Quarkus properties from previous build
```

quarkus-build-caching-extension/src/main/java/com/gradle/QuarkusBuildCache.java
```diff
@@ -100,12 +100,15 @@ private void configureQuarkusBuildGoal(MojoMetadataProvider.Context context, Qua
...
                 String baseDir = context.getProject().getBasedir().getAbsolutePath();
-                Properties quarkusCurrentProperties = QuarkusExtensionUtil.loadProperties(context.getProject().getBasedir().getAbsolutePath(), extensionConfiguration.getCurrentConfigFileName());
+                Properties quarkusPreviousProperties = QuarkusExtensionUtil.loadProperties(baseDir, extensionConfiguration.getDumpConfigFileName());
+
+                // Load Quarkus properties for current build
```
runningcode, line 107:
```suggestion
                // Load Quarkus properties from current build
```

runningcode (approved):
just some nits, otherwise looks great

PR #1255, opened 2024-06-20
Github actions access token migration

ribafish:
- Migrated GitHub actions gradle executions to access token auth 
- Migrated GitHub actions maven executions env var access key to develocity naming
- Replaced cmdline setting of dv server to develocity naming in github actions (`-Ddevelocity.url=<url>`)


jprinet (approved):
LGTM

PR #1280, opened 2024-07-01
Add quarkus-artifact.properties as an output of quarkus:build

gsmet:
Since the cache for quarkus:build has been fixed in the Quarkus CI, we are starting to see failures because the `target/quarkus-artifact.properties` is missing. This is because it's not part of the cache so whenever we have a cache hit, it's not restored.

----

When running ITs for native, we need quarkus-artifact.properties to be around so it needs to be part of the cache.

Also make it an input of Failsafe.
This file contains a relative path to the binary to run and a few additional metadata such as the full GraalVM version. Whenever this version has changed, we need to run the tests again so it makes sense to make it an input.

Note that we will need https://github.com/quarkusio/quarkus/pull/41587 to be in so that quarkus-artifact.properties doesn't contain a timestamp.


gsmet:
Note that we will need a release of the extension and an update in Quarkus once this is included as it's in the way of our CI. Thanks!

PR #1314, opened 2024-07-25
Allow cross OS cache hits when using in-container build strategy

jprinet:
Remove the OS characteristics as goal inputs when the in-container build strategy is configured.
This mimics the Java Compile goal configuration (OS details are not an input),

### DIFF
--- a/quarkus-build-caching-extension/src/main/java/com/gradle/QuarkusBuildCache.java
+++ b/quarkus-build-caching-extension/src/main/java/com/gradle/QuarkusBuildCache.java
@@ -161,7 +161,9 @@ final class QuarkusBuildCache {
 
         if (!quarkusPropertiesCopy.isEmpty()) {
             LOGGER.info(QuarkusExtensionUtil.getLogMessage("Quarkus properties have changed"));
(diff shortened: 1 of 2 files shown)


runningcode (approved):

quarkus-build-caching-extension/src/main/java/com/gradle/QuarkusBuildCache.java
```diff
@@ -218,7 +230,10 @@ private String getJarPackageType(Properties quarkusCurrentProperties) {
 
     private void configureInputs(MojoMetadataProvider.Context context, QuarkusExtensionConfiguration extensionConfiguration, Properties quarkusCurrentProperties) {
         context.inputs(inputs -> {
-            addOsInputs(inputs);
+            if (!isInContainerWithFixedImageBuild(quarkusCurrentProperties, false)) {
```
runningcode, line 233: just to confirm, we still need this for the non-container strategy because it affects the outcome?
why doesn't it affect the outcome on the in container strategy?

jprinet: Yes to your first question.
Native executable compilation depends on system libraries, we are assuming that building inside the same image would provide the same system libraries.

PR #1315, opened 2024-07-29
Convention sample for single build Gradle plugin and Maven extension is added

erichaagdev:
This PR adds a working sample of a single build containing both a Convention Develocity Gradle plugin & Convention Develocity Maven extension, where common Develocity configuration is shared between the two. 

The sample should feel very similar to the existing [Convention Develocity Gradle plugin](https://github.com/gradle/develocity-build-config-samples/tree/main/convention-develocity-gradle-plugin) and [Convention Develocity Maven extension](https://github.com/gradle/develocity-build-config-samples/tree/main/convention-develocity-maven-extension) that exist already in this repository.

The structure of the sample is as follows (also visible in the README of the sample):

* `convention-develocity-gradle-plugin` - Contains the convention Gradle plugin
* `convention-develocity-maven-extension` - Contains the convention Maven extension
* `convention-develocity-shared` - Contains the convention logic shared between build tools
* `examples` - Contains example builds that apply the convention Gradle plugin or Maven extension for different Gradle and Maven versions
    * `gradle_5` - Applies the convention Gradle plugin on a Gradle 5 build
    * `gradle_6` - Applies the convention Gradle plugin on a Gradle 6 build
    * `gradle_6.9_and_later` - Applies the convention Gradle plugin on a Gradle 6.9 and later build
    * `maven_3` - Applies the convention Maven extension on a Maven 3 build

### Testing

Just as was added for the other convention samples, a GitHub Actions workflow has been added that publishes the conventions locally and runs the example builds. The workflow tests both the Gradle and Maven conventions.

### Deviations

Everything in this PR matches what was proposed in the specification, except for shadowing `convention-develocity-shared` in `convention-develocity-gradle-plugin` and `convention-develocity-maven-extension`. While possible, this added a lot of additional build logic that I felt detracted from the point of the sample. As this is only a sample, shadowing all components to a single jar can still be done if desired by the user (as could be done for the other samples too).


.github/workflows/convention-develocity-maven-extension-verification.yml
```diff
@@ -62,11 +62,9 @@ jobs:
       - name: Set Maven version
         if: ${{ matrix.versions.version != '(Current)' }}
         working-directory: convention-develocity-maven-extension/examples/maven_${{ matrix.versions.sample }}
-        run: ./mvnw wrapper:wrapper -Dmaven=${{ matrix.versions.version }} -Dgradle.enterprise.url=https://ge.solutions-team.gradle.com
-        env:
-          DEVELOCITY_ACCESS_KEY: ${{ secrets.GE_SOLUTIONS_ACCESS_TOKEN }}
+        run: ./mvnw wrapper:wrapper -Dmaven=${{ matrix.versions.version }} -Ddevelocity.scan.disabled
```
clayburn, line 65: Why do we disable this scan?

convention-develocity-gradle-plugin/README.md
```diff
@@ -66,7 +66,7 @@ cd plugins/gradle-5-or-newer
 ./gradlew publishToMavenLocal
 ```
 
-Once you have published the plugins, you can run the four example builds under `examples`:
+Once you have published the plugins, you can run the five example builds under `examples`:
```
clayburn, line 69: Might be easier just to say `run the example builds` so these don't go out of sync.

.github/workflows/convention-develocity-shared-verification.yml
```diff
@@ -0,0 +1,123 @@
...
+        id: build
+        working-directory: convention-develocity-shared/examples/gradle_${{ matrix.versions.sample }}
+        run: ./gradlew build -Ddevelocity.url=https://ge.solutions-team.gradle.com
+        env:
+          DEVELOCITY_ACCESS_KEY: ${{ secrets.DV_SOLUTIONS_ACCESS_KEY }}
+      - name: Verify Build Scan published
+        if: ${{ !steps.build.outputs.build-scan-url }}
+        run: echo "::error ::No Build Scan published"; exit 1
+
+  verification:
```
clayburn, line 87: maybe should be `verification_maven` to match `verification_gradle` above

.github/workflows/convention-develocity-maven-extension-verification.yml
erichaagdev: It's only setting the version of the wrapper so I don't think it's necessary. The Gradle workflow [already does this](https://github.com/gradle/develocity-build-config-samples/blob/0ad93f704f3ba006e363132b3e7fe8d8ef075792/.github/workflows/convention-develocity-gradle-plugin-verification.yml#L80-L81) so this just brings it in line with that. 

This has nothing to do with the sample. I only noticed it as an inconsistency while creating the new workflow and wanted to fix it.

convention-develocity-gradle-plugin/README.md
erichaagdev: Good point.

.github/workflows/convention-develocity-shared-verification.yml
erichaagdev: Yes, it should be. Good catch.

convention-develocity-gradle-plugin/README.md
erichaagdev: Fixed in 058f77ec751600a6cea0438ee6825b5121acf44c

.github/workflows/convention-develocity-shared-verification.yml
erichaagdev: Fixed in 230d39a99170dfbab692f4812392e3ef51c7ca4f

.github/workflows/convention-develocity-maven-extension-verification.yml
clayburn: Works for me 👍

convention-develocity-shared/convention-develocity-shared/src/main/java/com/myorg/DevelocityConventions.java
```diff
@@ -0,0 +1,33 @@
...
+
+import com.myorg.configurable.BuildCacheConfigurable;
+import com.myorg.configurable.BuildScanConfigurable;
+import com.myorg.configurable.DevelocityConfigurable;
+
+final class DevelocityConventions {
+
+    void configureDevelocity(DevelocityConfigurable develocity) {
+        // CHANGE ME: Apply your Develocity configuration here
+        develocity.setServer("https://develocity-samples.gradle.com");
```
jprinet, line 11: Not sure if it would make sense to offer a way to use this directly without an obligation to change the code / recompile / redeploy (although it is not too involved).

I am guessing a lib which would contain a configuration file and ship the convention-develocity which would load data from this configuration file if present could be published instead.

jprinet (approved):
Great job Eric, all clear!

I left a comment with a fancy idea to try to ease consumption of the lib. Not sure that would help a lot though.

convention-develocity-shared/convention-develocity-shared/src/main/java/com/myorg/DevelocityConventions.java
jprinet: My point is that this reference could be even more practical if you wouldn't require to fork/change the Java code to use it.
A customer could create its DV custom extension as a lib containing a configuration file and depending on the convention-develocity lib. 
This was just a thought though.

erichaagdev: Ah, I understand what you mean now.

It's an interesting thought, but whether it's updating a properties file, using a library, or copying/pasting the contents of this sample, they must create a new repository to tailor it to their needs. The latter gives complete control over the behavior so I think we should stick with that.

Where I think we can improve is to externalize the "boilerplate" adapter code. As it is right now, it creates a lot of noise in the sample.

.github/workflows/convention-develocity-maven-extension-verification.yml
erichaagdev: FYI – I decided to break this out to a separate PR in case this PR takes a while to be merged. https://github.com/gradle/develocity-build-config-samples/pull/1318

clayburn (approved):

PR #1318, opened 2024-08-02
Workflows are improved

erichaagdev:
This PR contains several very minor fixes & improvements to the workflows in this repository. See each commit for details.

### DIFF
--- a/.github/workflows/convention-develocity-gradle-plugin-verification.yml
+++ b/.github/workflows/convention-develocity-gradle-plugin-verification.yml
@@ -80,13 +80,13 @@ jobs:
           ./gradlew wrapper --gradle-version=${{ matrix.versions.version }} --no-scan
           ./gradlew wrapper --gradle-version=${{ matrix.versions.version }} --no-scan
         env:
-          DEVELOCITY_ACCESS_KEY: ${{ secrets.GE_SOLUTIONS_ACCESS_TOKEN }}
+          DEVELOCITY_ACCESS_KEY: ${{ secrets.DV_SOLUTIONS_ACCESS_KEY }}
       - name: Verify example build
         id: build
         working-directory: convention-develocity-gradle-plugin/examples/gradle_${{ matrix.versions.sample }}
         run: ./gradlew build -Ddevelocity.url=https://ge.solutions-team.gradle.com
         env:
-          DEVELOCITY_ACCESS_KEY: ${{ secrets.GE_SOLUTIONS_ACCESS_TOKEN }}
+          DEVELOCITY_ACCESS_KEY: ${{ secrets.DV_SOLUTIONS_ACCESS_KEY }}
       - name: Verify Build Scan published
         if: ${{ !steps.build.outputs.build-scan-url }}
         run: echo "::error ::No Build Scan published"; exit 1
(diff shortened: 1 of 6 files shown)


.github/workflows/convention-develocity-gradle-plugin-verification.yml
```diff
@@ -3,10 +3,10 @@ name: Verify Convention Develocity Gradle Plugin
 on:
   push:
     branches: [ main ]
-    paths: [ 'convention-develocity-gradle-plugin/**' ]
+    paths: [ 'convention-develocity-gradle-plugin/**', '.github/workflows/convention-develocity-gradle-plugin-verification.yml' ]
```
jprinet, line 6: Not sure we should be too fine grained here, as a rename is likely to remain unnoticed.
I'd rather go with prefixed patterns

jprinet (approved):
LGTM although I feel we should relax a bit the run condition

.github/workflows/convention-develocity-gradle-plugin-verification.yml
erichaagdev: That's true, but workflows are unlikely to change often and when they do, the impact of missing this is minimal. How would you configure the prefixed patterns?

jprinet: Maybe we could afford '.github/workflows/**'

bigdaz: I agree with Jerome: making it simpler probably outweighs any advantage to only running the changed workflow.

erichaagdev: Do you know if GitHub exposes a variable for "path to current workflow" and if it could be used within an `on.push.paths:` condition? I'd really like to avoid running all workflows with `.github/workflows/**` when any workflow has changed. It would cause an explosion of unnecessary workflow runs which is where we used to be before adding these `on:` conditions in the first place.

jprinet: Not afaik

clayburn: @erichaagdev - I can't link as directly to it as I would like, but I see a `github.workflow_ref` that contains (as part of a longer string) the workflow file name: https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/contexts#github-context

Would this solve your problem?

erichaagdev: I don't think so. We need the path to the workflow file exactly. Even if we had that though, I'm not sure it would work. Will need to test it.

erichaagdev: It turns out the `github` context isn't available in `on.push.paths`: https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/contexts#context-availability

clayburn: In that case, I would lean towards Jerome's solution above (`.github/workflows/**`).

erichaagdev: Updated in b2157306b2e6ebf3746c0fb1ce9c0b62029fa0a0

clayburn (approved):

PR #1326, opened 2024-08-06
Added testing kts scripts for gradle data capturing samples

ribafish:
Also added testing for capturing max parallel forks, as that was missing in the GHA workflow.


ribafish:
Missed that this was already implemented, just not as part of a matrix, closing

PR #1331, opened 2024-08-07
Migrated a gradle groovy sample to develocity plugin api

ribafish:
Migrated a missed gradle groovy sample to develocity plugin api

### DIFF
--- a/build-data-capturing-gradle-samples/capture-test-pts-support/gradle-test-pts-support.gradle
+++ b/build-data-capturing-gradle-samples/capture-test-pts-support/gradle-test-pts-support.gradle
@@ -12,11 +12,11 @@ import java.util.stream.Collectors
  * adding a flag as custom value.
  */
 
-def buildScanApi = project.extensions.findByName('develocity').getBuildScan();
-if (!buildScanApi) {
+def develocityApi = project.extensions.findByName('develocity')
+if (!develocityApi) {
     return
 }
-def capture = new Capture(buildScanApi, gradle.rootProject.logger)
+def capture = new Capture(develocityApi.getBuildScan(), gradle.rootProject.logger)
 allprojects {
     tasks.withType(Test).configureEach { t ->
         doFirst {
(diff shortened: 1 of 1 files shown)


build-data-capturing-gradle-samples/capture-test-pts-support/gradle-test-pts-support.gradle
```diff
@@ -12,7 +12,7 @@ import java.util.stream.Collectors
  * adding a flag as custom value.
  */
 
-def buildScanApi = project.extensions.findByName('buildScan')
+def buildScanApi = project.extensions.findByName('develocity').getBuildScan();
```
erichaagdev, line 15: This is different than how we do it in all other samples.

https://github.com/gradle/develocity-build-config-samples/blob/b84ea9bf43101c3c524bbe3e8a04e3a28f83fd75/build-data-capturing-gradle-samples/capture-os-processes/gradle-os-processes.gradle#L9-L12

I believe this will throw a NPE if the `develocity` extension doesn't exist, though it's unlikely you would be applying this with the Develocity Gradle plugin applied.

ribafish: Yes, I know it's different, but I'm not sure how to do it differently in groovy, as doing what was there before get's you an error: `Could not find matching constructor for: Capture(com.gradle.enterprise.gradleplugin.internal.extension.b_Decorated, org.gradle.internal.logging.slf4j.OutputEventListenerBackedLogger)` - [scan](https://ge.solutions-team.gradle.com/s/eb7nko4b636eo). I'll handle the develocity being null though, missed that.

erichaagdev: I see you pushed some changes. Does it fix the error you're describing?

erichaagdev: Okay so the script is still broken?

ribafish: No, the way I opened the PR is not broken - I wouldn't open a PR otherwise (it also shouldn't pass the tests if it were broken) :)

erichaagdev (approved):

PR #1384, opened 2024-08-26
Convention samples access environment in configuration cache compatible way

erichaagdev:
This changes the Gradle convention sample to read environment variables in such a way that changes are detected when configuration cache is enabled in all versions of Gradle.

The shared convention sample was also updated to abstract how environment variables (and system properties) are read between build tools and also have received the same configuration cache compatible changes for the Gradle plugin.

Currently, the only use case for reading environment variables is to know if the build is running in CI which will typically not change between build invocations under normal circumstances. However, users forking the sample may desire reading other environment variables or system properties that are more likely to change between build invocations, which is generally pretty common. To encourage using best practices and avoid having to explain the nuances of configuration cache behavior between Gradle versions _if_ they decide to extend the sample, it was determined the easiest path forward is to just implement it in the sample so it works out of the box.

This is [the same environment variable & system property logic used in CCUD](https://github.com/gradle/common-custom-user-data-gradle-plugin/blob/461581adcd86913a754d09c57c80f8875f93ea42/src/main/java/com/gradle/Utils.java#L49-L54).

### DIFF
--- a/convention-develocity-gradle-plugin/plugins/gradle-5-or-newer/src/main/java/com/myorg/ConventionDevelocityGradlePlugin.java
+++ b/convention-develocity-gradle-plugin/plugins/gradle-5-or-newer/src/main/java/com/myorg/ConventionDevelocityGradlePlugin.java
@@ -9,14 +9,26 @@ import org.gradle.api.GradleException;
 import org.gradle.api.Plugin;
 import org.gradle.api.Project;
 import org.gradle.api.initialization.Settings;
+import org.gradle.api.provider.Provider;
+import org.gradle.api.provider.ProviderFactory;
 import org.gradle.caching.configuration.BuildCacheConfiguration;
 import org.gradle.util.GradleVersion;
+
+import javax.inject.Inject;
+import java.util.Optional;
 
 /**
  * An example Gradle plugin for enabling and configuring Develocity features for
  * Gradle versions 5.x and higher.
  */
 public class ConventionDevelocityGradlePlugin implements Plugin<Object> {
+
+    private final ProviderFactory providers;
+
+    @Inject
+    public ConventionDevelocityGradlePlugin(ProviderFactory providers) {
+        this.providers = providers;
+    }
 
     @Override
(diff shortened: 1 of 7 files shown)


convention-develocity-gradle-plugin/plugins/gradle-5-or-newer/src/main/java/com/myorg/ConventionDevelocityGradlePlugin.java
```diff
@@ -83,9 +95,27 @@ private void configureBuildCache(BuildCacheConfiguration buildCache, DevelocityC
         });
     }
 
-    private static boolean isCi() {
+    private boolean isCi() {
         // CHANGE ME: Apply your environment detection logic here
-        return System.getenv().containsKey("CI");
+        return environmentVariable("CI").isPresent();
```
etiennestuder, line 100: Do containsKey("CI") and isPresent() have the same semantics, particularly when the key is present but the value is null or the empty string?

etiennestuder (approved):

clayburn (approved):

convention-develocity-gradle-plugin/plugins/gradle-5-or-newer/src/main/java/com/myorg/ConventionDevelocityGradlePlugin.java
erichaagdev: Yes. If defined but with an empty string, `System.getenv("CI")` will return an empty string. An empty string is still considered valued and therefore "present" in the eyes of `Optional`.

`System.getenv("CI")` will only return `null` if the `CI` environment variable is not defined in the system. 

This reflects the semantics of `containsKey`.

etiennestuder: Great. Thanks for the confirmation.

PR #1394, opened 2024-08-29
Maven listener isn't called in shared sample

erichaagdev:
This was missed when creating the shared build. It meant the listener was never called. I never encountered it because I always ran with `-Ddevelocity.url=https://ge.solutions-team.gradle.com`.

### DIFF
--- a/.github/workflows/convention-develocity-shared-verification.yml
+++ b/.github/workflows/convention-develocity-shared-verification.yml
@@ -23,6 +23,9 @@ jobs:
           distribution: 'temurin'
       - name: Set up Gradle
         uses: gradle/actions/setup-gradle@v4
+      - name: Configure Develocity Solutions instance
+        working-directory: convention-develocity-shared/convention-develocity-common/src/main/java/com/myorg
+        run: sed -i 's/develocity-samples.gradle.com/ge.solutions-team.gradle.com/g' DevelocityConventions.java
       - name: Build with Gradle
         working-directory: convention-develocity-shared
         run: ./gradlew build publishToMavenLocal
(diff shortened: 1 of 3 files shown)


etiennestuder:
LGTM, but I cannot assess the overall impact. I will rely on @erichaagdev having verified it works.

etiennestuder:
@erichaagdev do we need to do a similar change for the shared-plugin-extension ?

etiennestuder (approved):

etiennestuder:
> LGTM, but I cannot assess the overall impact. I will rely on @erichaagdev having verified it works.

I meant the non-shared convention extension.

jprinet (approved):
LGTM, I could test that locally!
Great fast work Eric 👍

erichaagdev:
@etiennestuder it only is a problem in the shared sample. 

In the standalone Maven sample, it is built with Maven and the `plexus-component-metadata` generates the `component.xml` file as part of the build.

This isn't the case in the shared sample, because it is built with Gradle. 

Now that I think about it, this is actually redundant:

https://github.com/gradle/develocity-build-config-samples/blob/0ceecfbbb047c1664c5c35add42fcecc1f458692/convention-develocity-shared/convention-develocity-maven-extension/src/main/java/com/myorg/ConventionDevelocityMavenExtensionListener.java#L13-L17

We don't include it in [the extension built by Gradle in the BVS repo](https://github.com/gradle/gradle-enterprise-build-validation-scripts/blob/73106aee0e3c09d794e04a7e15dd47d099823ad3/components/configure-gradle-enterprise-maven-extension/src/main/java/com/gradle/ConfigureDevelocity.java#L10).

PR #1396, opened 2024-09-04
Include LICENSE file in published Jar

jprinet:
This PR adds a `LICENSE` file to the root of the project, it is then shipped in the published Jar

Note:
there is no `NOTICE` file because the extension does not have third-party dependencies

### DIFF
--- a/quarkus-build-caching-extension/pom.xml
+++ b/quarkus-build-caching-extension/pom.xml
@@ -145,6 +145,9 @@
         </plugins>
         <resources>
             <resource>
+                <directory>${project.basedir}/src/main/resources</directory>
+            </resource>
+            <resource>
                 <targetPath>META-INF</targetPath>
                 <directory>${basedir}</directory>
                 <includes>
--- a/quarkus-build-caching-extension/pom.xml
+++ b/quarkus-build-caching-extension/pom.xml
@@ -148,7 +148,6 @@
                 <directory>${project.basedir}/src/main/resources</directory>
             </resource>
             <resource>
-                <targetPath>META-INF</targetPath>
                 <directory>${project.basedir}</directory>
                 <includes>
                     <include>LICENSE</include>
--- a/quarkus-build-caching-extension/pom.xml
+++ b/quarkus-build-caching-extension/pom.xml
@@ -143,6 +143,15 @@
                 </executions>
             </plugin>
         </plugins>
+        <resources>
+            <resource>
+                <targetPath>META-INF</targetPath>
+                <directory>${basedir}</directory>
+                <includes>
+                    <include>LICENSE</include>
+                </includes>
+            </resource>
+        </resources>
     </build>
 
     <profiles>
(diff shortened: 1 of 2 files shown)


quarkus-build-caching-extension/pom.xml
```diff
@@ -143,6 +143,15 @@
                 </executions>
             </plugin>
         </plugins>
+        <resources>
+            <resource>
+                <targetPath>META-INF</targetPath>
```
erichaagdev, line 148: Will this include the license at `META-INF/LICENSE` in the JAR? We want it at the top level of the JAR.

quarkus-build-caching-extension/pom.xml
```diff
@@ -143,6 +143,15 @@
                 </executions>
             </plugin>
         </plugins>
+        <resources>
```
erichaagdev, line 146: You'll notice in the CCUD Maven PR I redefined `src/main/resources` as explained [in this comment](https://github.com/gradle/common-custom-user-data-maven-extension/pull/241#discussion_r1743820327).

I think it would be nice to do the same here too for consistency, unless you or Etienne don't agree with my reasoning.

quarkus-build-caching-extension/pom.xml
```diff
@@ -143,6 +143,15 @@
...
+            <resource>
+                <targetPath>META-INF</targetPath>
+                <directory>${basedir}</directory>
```
erichaagdev, line 149: Can you make it `project.basedir` here? We do that in other places in this POM, e.g., `project.build.directory`, and that's how I did it in CCUD Maven.

quarkus-build-caching-extension/pom.xml
jprinet: Yes it will include the `LICENSE` file in the `JAR` but in the `META-INF` folder.
Why do you want it at the top-level?

quarkus-build-caching-extension/pom.xml
jprinet: Sure thing

quarkus-build-caching-extension/pom.xml
jprinet: Sure thing

quarkus-build-caching-extension/pom.xml
jprinet: Adjusted to root level for consistency

erichaagdev (approved):
Approved, thanks! But let's wait to merge until @etiennestuder has approved CCUD Maven. That way any additional changes are requested they can be made here without a new PR.

quarkus-build-caching-extension/pom.xml
```diff
@@ -143,6 +143,17 @@
                 </executions>
             </plugin>
         </plugins>
+        <resources>
+            <resource>
+                <directory>${project.basedir}/src/main/resources</directory>
+            </resource>
+            <resource>
```
etiennestuder, line 150: I would take this (src/main/resources) out here, too.

etiennestuder (changes requested):

quarkus-build-caching-extension/pom.xml
erichaagdev: As discussed in the CCUD Maven PR, we will leave it in.

PR #1979, opened 2025-08-25
Improve Gradle verification workflows

ribafish:
- Simplified the workflow for shared plugin verification for Gradle - builds now rely on the version set by setup-gradle action
- Renamed matrix keys for clarity

### DIFF
--- a/.github/workflows/convention-develocity-gradle-plugin-verification.yml
+++ b/.github/workflows/convention-develocity-gradle-plugin-verification.yml
@@ -49,38 +49,38 @@ jobs:
         # Gradle 4 is not tested because it does not support publishing to authenticated servers.
         versions:
           - sample: '5'
-            version: '5.0'
-            java: '8'
+            gradle-version: '5.0'
+            java-version: '8'
           - sample: '6'
-            version: '6.0.1'
-            java: '8'
+            gradle-version: '6.0.1'
+            java-version: '8'
           - sample: '6.9_and_later'
-            version: '6.9.4'
-            java: '8'
+            gradle-version: '6.9.4'
+            java-version: '8'
           - sample: '6.9_and_later'
-            version: '7.0.2'
-            java: '11'
+            gradle-version: '7.0.2'
+            java-version: '11'
           - sample: '6.9_and_later'
-            version: '8.0.2'
-            java: '17'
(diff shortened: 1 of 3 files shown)


.github/workflows/convention-develocity-gradle-plugin-verification.yml
```diff
@@ -69,6 +69,9 @@ jobs:
           - sample: '6.9_and_later'
             version: 'wrapper'
             java: '21'
+          - sample: '6.9_and_later'
+            version: 'release-candidate'
+            java: '21'
```
erichaagdev, line 74: What do you think about renaming these to `gradle-version` and `java-version`, to be more explicit? If you do that, I would also apply a similar change to Maven.

.github/workflows/convention-develocity-gradle-plugin-verification.yml
```diff
@@ -69,6 +69,9 @@ jobs:
           - sample: '6.9_and_later'
             version: 'wrapper'
             java: '21'
+          - sample: '6.9_and_later'
+            version: 'release-candidate'
```
erichaagdev, line 73: I like the idea of testing RC versions, but I don't know if this is the right way to do it. A better way would be to have a nightly job that runs against the RC version. As it is now, RCs will only be tested if an unrelated PR is opened. For example, Dependabot creates a PR to upgrade a dependency, and the build fails because of some unrelated incompatibility with a Gradle RC.

erichaagdev (commented):
Just a few comments.

.github/workflows/convention-develocity-gradle-plugin-verification.yml
ribafish: Done

.github/workflows/convention-develocity-gradle-plugin-verification.yml
ribafish: Yes, that's true. Do you think we should simply add a cron trigger to the verification workflows, or do you propose creating a new one to just run against the RC. Bear in mind that we'd be doing this only for Gradle.

ribafish: Deferred to a later PR (after a discussion).

PR #2142, opened 2026-01-13
DO NOT MERGE: Add GitHub custom property as custom data to Build Scans

erichaagdev:
Not intended to be merged.

### DIFF
--- a/convention-develocity-shared/convention-develocity-common/src/main/java/com/myorg/DevelocityConventions.java
+++ b/convention-develocity-shared/convention-develocity-common/src/main/java/com/myorg/DevelocityConventions.java
@@ -5,6 +5,8 @@ import com.myorg.configurable.BuildScanConfigurable;
 import com.myorg.configurable.DevelocityConfigurable;
 import com.myorg.configurable.ExecutionContext;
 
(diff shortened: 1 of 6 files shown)


convention-develocity-shared/convention-develocity-common/src/main/java/com/myorg/DevelocityConventions.java
```diff
@@ -16,14 +16,14 @@ final class DevelocityConventions {
     void configureDevelocity(DevelocityConfigurable develocity) {
         // CHANGE ME: Apply your Develocity configuration here
         develocity.setServer("https://develocity-samples.gradle.com");
-        configureBuildScan(develocity.getBuildScan());
+        configureBuildScan(develocity, develocity.getBuildScan());
         configureBuildCache(develocity.getBuildCache());
     }
 
-    private void configureBuildScan(BuildScanConfigurable buildScan) {
+    private void configureBuildScan(DevelocityConfigurable develocity, BuildScanConfigurable buildScan) {
```
clayburn: Technically you could just pass DevelocityConfigurable here and get BuildScanConfigurable from it.

PR #2241, opened 2026-04-06
Remove licensing notice for using `gradle-actions-caching`

clayburn:
See [here](https://github.com/gradle/develocity-build-config-samples/actions/runs/24019341011/attempts/1#summary-70044920139) for an example


.github/workflows/convention-develocity-gradle-plugin-verification-nightly.yml
```diff
@@ -30,6 +30,8 @@ jobs:
           develocity-url: 'https://ge.solutions-team.gradle.com'
           develocity-plugin-version: '4.2.1'
           develocity-access-key: ${{ secrets.DV_SOLUTIONS_ACCESS_KEY }}
+          build-scan-terms-of-use-url: 'https://gradle.com/terms-of-service'
+          build-scan-terms-of-use-agree: 'yes'
```
bigdaz, line 34: This isn't required here, or anywhere that `develocity-access-key` is set.

PR #2261, opened 2026-05-20
Add sample for caching maven-checkstyle-plugin check goal

jprinet:
## Summary
- Let's not merge this one for the moment - the early results show little gain, which could mean negative savings in case of remote cache hit

- New `build-caching-maven-samples/checkstyle-project/` recipe demonstrating how to make the `check` goal of the `maven-checkstyle-plugin` cacheable via the Develocity Maven extension.
- Declares inputs for source/test source directories and resources, observable plugin properties, and the `${project.build.directory}/checkstyle-result.xml` output. Inline comments document the known limitations (config/header/suppressions/properties files referenced from the classpath are not tracked as content inputs; `xrefLocation`/`outputXrefLocation` are ignored — same caveat pattern used in the existing PMD/SpotBugs samples).
- Wired the new module into `build-caching-maven-samples/pom.xml` and added a Checkstyle entry to the Maven samples README.

## Test plan
- [x] `./mvnw -pl checkstyle-project clean verify` succeeds on a clean target.
- [x] Re-running with the local cache enabled produces a cache hit on `checkstyle:3.6.0:check` (verified via build scan: https://ge.solutions-team.gradle.com/s/7j3wldpksyvfw).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/build-caching-maven-samples/README.md
+++ b/build-caching-maven-samples/README.md
@@ -12,6 +12,7 @@ The following Maven plugins are demonstrated:
 - AsciiDoctor: `org.asciidoctor:asciidoctor-maven-plugin`
 - AspectJ: `org.codehaus.mojo:aspectj-maven-plugin`
 - Avro (Apache): `org.apache.avro:avro-maven-plugin`
+- Checkstyle: `org.apache.maven.plugins:maven-checkstyle-plugin`
 - Clojure: `com.theoryinpractise:clojure-maven-plugin`
 - Duplicate Finder Plugin: `org.basepom.maven:duplicate-finder-maven-plugin`
 - Enforcer: `org.apache.maven.plugins:maven-enforcer-plugin`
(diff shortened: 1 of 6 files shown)


jprinet:
Validated the corrected recipe by porting it into a fresh clone of prestodb/presto and running `./mvnw -pl presto-bytecode -DskipTests clean validate` twice. Second run produced a cache hit on `checkstyle:3.6.0:check`: https://ge.solutions-team.gradle.com/s/n3hvip2jba6wc

The original recipe in this PR silently broke caching on any project with a resources directory (ClassCastException Resource → File) and on any project where the plugin extension auto-classified `outputFile`/`useFile`/`rulesFiles`/`cacheFile` and `${checkstyle.*}` user-expression properties. This commit fixes both. The sample also now sets `includeTestSourceDirectory=true` so the `testSourceDirectories` fileSet is actually exercised (otherwise the extension auto-ignores it).

ribafish (approved):
LGTM

PR #2279, opened 2026-06-15
Skip spring-cloud-contract 5.0.3 in Renovate

erichaagdev:
5.0.3 breaks Maven build-caching verification (HttpClient `AuthCache` ClassRealm conflict). Bumped in #2271, rolled back in #2272, then re-proposed by Renovate in #2274.

`allowedVersions: [5.0.4,)` floors the version at 5.0.4 to skip 5.0.3; 5.0.4+ is allowed through when released.

Supersedes #2278 (auto-closed by GitHub after a force-push left its head with no common ancestor).

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -98,6 +98,17 @@
       "allowedVersions": "<2.0.0"
     },
     {
+      // spring-cloud-contract 5.0.3 breaks Maven build-caching verification
+      // (HttpClient AuthCache ClassRealm conflict); bumped in #2271, rolled back
+      // in #2272. Floor at 5.0.4 to skip the broken 5.0.3; 5.0.4+ allowed when released.
+      "matchManagers": ["maven"],
+      "matchPackageNames": [
+        "org.springframework.cloud:spring-cloud-starter-contract-verifier",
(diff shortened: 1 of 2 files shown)


jprinet:
The problem actually got fixed on Maven side https://github.com/apache/maven-resolver/issues/1919
I don't think 5.0.4 will ship any fix to this, holding off on the upgrade may help us to get a fixed Maven, though

jprinet (approved):

erichaagdev:
Good call. Since the fix is upstream and 5.0.4 likely won't carry it, I dropped the spring-cloud-contract floor and instead pinned `maven-resolver-transport-http` to 1.9.25 in the plugin's own dependencies, which is where the buggy 1.9.27 actually enters (its `extensions=true` realm, not Maven core). The sample stays on 5.0.3 and the build stays green.

The Renovate rule now skips transport-http 1.9.26/1.9.27 and will auto-bump to 1.9.28 once it ships, at which point the pin can be removed.

Verified locally against a cold repo: 5.0.3 fails 3/3 unpinned with the `BasicAuthCache` cast, succeeds 3/3 with the pin (generateTests runs).

jprinet (approved):
Great 👍

PR #2289, opened 2026-06-26
Document how to condition build cache push on credentials presence (fork PR support)

jprinet:
When CI builds are triggered from forked repositories, secrets are not passed to the runner, causing cache store attempts to fail with HTTP 403.

This PR adds a commented-out alternative to the `storeEnabled` / `push` / `isPush` settings in all four sample configs, showing how to also condition on the access key being present for teams accepting contributions through forks.


jprinet:
We decided to keep the samples as straightforward as possible

